PR #2, opened 2018-11-02
Add continuous integration using CircleCI

OndrejMarsalek:
This is a basic setup of continuous integration on CircleCI. I tested it on my fork and it runs and the tests (that are not skipped) pass.

I also made some minor changes to the tests to avoid the dependence on Nose test (that we're not using anymore) and to skip the manual build tests if LaTeX is not found.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,45 @@
+# Python CircleCI 2.0 configuration file
+#
+# Check https://circleci.com/docs/2.0/language-python/ for more details
+#
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/python:2.7-stretch
+
+    working_directory: ~/repo
+
+    steps:
+      - checkout
+
+      # Download and cache dependencies
+      - restore_cache:
+          keys:
+          - v1-dependencies-{{ checksum "requirements.txt" }}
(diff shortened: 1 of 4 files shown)


OndrejMarsalek:
As I already enabled it in the CircleCI interface, you can see that the check ran for this branch. I will also add this to the branch protection rules - we want tests to pass for a PR before it can be merged.

ceriottm:
Is there any way/would it be appropriate to also run regtests with CI? Currently the unit tests we have are pretty useless, while the regtests are relatively thorough

OndrejMarsalek:
Technically, nothing stands in the way, you have a Docker container and you can go to town with what exactly your tests are. In practice, it will be more work to set it up, as it should probably be a separate "check" (in the lingo of GitHub CI), and it needs more setting up in the Docker image. Speed is another potential concern, but what do I know, maybe it will be fine.

Personally, I think this should be taken as an opportunity to improve the "regular" tests. Having integrated tests is useful, but good "unit tests" will tell you more specifically what's broken, and usually faster. My hope is that eventually we can get decent test coverage and require that it does decrease as part of the check.

Can we start with this for now and remember that we want to eventually run regtests automatically too?

PR #4, opened 2018-11-05
A simple change to the poll routine can reduce latency by 2x

ceriottm:
An incredibly simple change to the main poll loop of forcefields base can reduce up to 2x the overhead when dealing with super-cheap potentials, e.g. plumed. This is quite a game changer so I wanted to include it quickly in master while we finalize the remd-metad branch.

### DIFF
--- a/ipi/engine/forcefields.py
+++ b/ipi/engine/forcefields.py
@@ -128,16 +128,10 @@ class ForceField(dobject):
 
         # Indexes come from input in a per atom basis and we need to make a per atom-coordinate basis
         # Reformat indexes for full system (default) or piece of system
-#        fullat=True
         if self.active[0] == -1:
             activehere = np.array([i for i in range(len(pbcpos))])
         else:
             activehere = np.array([[3 * n, 3 * n + 1, 3 * n + 2] for n in self.active])
-
-#           fullat=False
-#
-#        if (self.active[0]!=-1 and fullat==False):
-#           temp=np.array([[3*n, 3*n+1, 3*n+2] for n in self.active])
 
         # Reassign active indexes in order to use them
         activehere = activehere.flatten()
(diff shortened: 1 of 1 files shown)


ipi/engine/forcefields.py
```diff
@@ -521,7 +521,7 @@ def __init__(self, latency=1.0e-3, name="", pars=None, dopbc=False, init_file=""
         self.charges = dstrip(myatoms.q) * 0.0
         self.masses = dstrip(myatoms.m)
         self.lastq = np.zeros(3 * self.natoms)
-
+
```
OndrejMarsalek: I recommend setting your editor to remove trailing whitespace on every save ;-)

OndrejMarsalek:
And as an admin you can override the pending check which is not active on this branch yet.

ceriottm:
so why isn't the check active? is there something I should do?

mahrossi:
Looks amazing to me - but just to clarify what did the trick? The new if/else to sleep only if there are no requests?

OndrejMarsalek:
@ceriottm, it looks like it is a separate setting in CircleCI to trigger builds for PRs from forks, and it was off by default. I changed it now. Can you try to trigger a build by another commit, say, removing that trailing whitespace?

OndrejMarsalek:
BTW, I think the "good first issue" label is meant for issues that are bugs/enhancements that you want to encourage junior people to pick up - something straightforward that just needs doing that is an opportunity to practice the whole workflow.

ipi/engine/forcefields.py
```diff
@@ -174,12 +168,16 @@ def queue(self, atoms, cell, reqid=-1):
     def poll(self):
         """Polls the forcefield object to check if it has finished."""
 
-        for r in self.requests:
-            if r["status"] == "Queued":
-                r["t_dispatched"] = time.time()
-                r["result"] = [0.0, np.zeros(len(r["pos"]), float), np.zeros((3, 3), float), ""]
-                r["status"] = "Done"
-                r["t_finished"] = time.time()
+        self._threadlock.acquire()
```
OndrejMarsalek: It looks like it would be nice to use a context manager here: https://docs.python.org/2.7/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement

OndrejMarsalek: OTOH, it might be better to do a full pass over all the `acquire` calls separately. Up to you.

PR #5, opened 2018-11-06
One pepper pass

OndrejMarsalek:
This brings the new repo in line with the currently configured autopep, which seems to be a good idea for a new beginning.

I would normally group this with something more substantial, but I also want to see that CircleCI now correctly triggers for PRs from forks.


OndrejMarsalek:
And CI did indeed run! Someone please review this and merge.

PR #9, opened 2018-11-29
Feat/ffquip

venkatkapil24:
Any potential implemented in QUIP can now be used in i-pi directly. Added examples for Si and an OH morse potential.


ipi/engine/forcefields.py
```diff
@@ -375,6 +375,96 @@ def evaluate(self, r):
...
+
+        # Initializes an atoms object and the interaction potential
+        self.atoms = quippy.Atoms(self.init_file)
+        self.pot = quippy.Potential(self.args_str, param_filename=self.param_file)
+       
+        # Initializes the conversion factors from i-pi to QUIP
+        self.len_conv = 0.529177 
+        self.energy_conv = 27.211386
+
+    def poll(self):
```
ceriottm: Make sure this is actually needed. Also after this is merged please take care of merging master into feat/fast_ff because I suspect this will interact with the new default strategy which is not to thread over direct calls

ipi/engine/forcefields.py
```diff
@@ -375,6 +375,96 @@ def evaluate(self, r):
...
+    """
+
+    def __init__(self, latency=1.0e-3, name="", pars=None, dopbc=True):
+        """Initialises QUIP.
+
+        Args:
+        pars: Mandatory dictionaru, giving the parameters needed by QUIP.
+        """
+        if quippy is None:
+            raise ImportError("QUIPPY import failed due to expection : " + str(e))
```
ceriottm: exception. learn to spell ^_^

ipi/engine/forcefields.py
```diff
@@ -375,6 +375,96 @@ def evaluate(self, r):
...
+        self.args_str = str(self.pars["args_str"])
+        self.param_file = str(self.pars["param_file"])
+
+        # Initializes an atoms object and the interaction potential
+        self.atoms = quippy.Atoms(self.init_file)
+        self.pot = quippy.Potential(self.args_str, param_filename=self.param_file)
+       
+        # Initializes the conversion factors from i-pi to QUIP
+        self.len_conv = 0.529177 
+        self.energy_conv = 27.211386
```
ceriottm: I think these are defined in utils.Constants. please reuse, or use units_to_internal so we're sure we don't end up with unsolvable 1e-5 discrepancies in the future.

ipi/inputs/forcefields.py
```diff
@@ -207,6 +207,37 @@ def fetch(self):
             raise ValueError("Negative timeout parameter specified.")
 
 
+class InputFFQUIP(InputForceField):
+
+    attribs = {}
+    attribs.update(InputForceField.attribs)
+
+    default_help = """A general QUIP interaction potential evaluator.
```
ceriottm: Use fields (see e.g. plumedff) to read these options, rather than stuffing everyting into params

ipi/inputs/simulation.py
```diff
@@ -153,6 +154,10 @@ def store(self, simul):
                     _iobj = iforcefields.InputFFLennardJones()
                     _iobj.store(_obj)
                     self.extra[_ii] = ("fflj", _iobj)
+                elif isinstance(_obj, eforcefields.FFQUIP):
```
ceriottm: This might also need some work once you merge with feat/fast_ff as I added some provisions to reuse InputXXX objects when storing the checkpoint file (saving like 0.5s/step in complex REMD simulations)

ipi/engine/forcefields.py
ceriottm: @venkatkapil24  don't forget to fix these.

venkatkapil24: Yeah, I am aware of this one!

ipi/engine/forcefields.py
```diff
@@ -375,6 +373,97 @@ def evaluate(self, r):
         r["status"] = "Done"
 
 
+try:
```
OndrejMarsalek: I think this should be at the top of the file, even if it's optional.

ipi/engine/forcefields.py
```diff
@@ -375,6 +373,97 @@ def evaluate(self, r):
...
+    import quippy
+except Exception as e:
+    quippy = None
+    quippy_exc = e
```
OndrejMarsalek: I don't think the rename is needed.

ipi/engine/forcefields.py
```diff
@@ -375,6 +373,97 @@ def evaluate(self, r):
...
+            Of the form {'atoms': atoms, 'cell': cell, 'pars': parameters,
+                         'status': status, 'result': result, 'id': bead id,
+                         'start': starting time}.
+    """
+
+    def __init__(self, init_file, args_str, param_file, latency=1.0e-3, name="", pars=None, dopbc=True):
+        """Initialises QUIP.
+
+        Args:
+        pars: Mandatory dictionaru, giving the parameters needed by QUIP.
```
OndrejMarsalek: Typo.

ipi/engine/forcefields.py
```diff
@@ -375,6 +373,97 @@ def evaluate(self, r):
...
+        """
+        if quippy is None:
+            raise ImportError("QUIPPY import failed due to exception : " + str(e))
```
OndrejMarsalek: `e` is the original variable, not the rename, so that's not needed. I would use `str(exc_quippy)` here and `except Exception as exc_quippy:` above.

OndrejMarsalek: I would also consider doing something other than `ImportError` here. This is the error the user will see they don't have quippy installed, so ideally it would not end up as an unhandled exception. Print error message and exit with a non-zero return code would be the expected behaviour in a case like this, I think.

PR #10, opened 2018-11-30
Feat/gleacf

venkatkapil24:
1) Cleaned up the tool.
2) Updated the examples and their READMEs.
3) Extended the tool to NMGLEs.
4) Also added an option to add the A and C matrices manually.


tools/py/gleacf.py
```diff
@@ -145,61 +145,83 @@ def ISRA(omega, ker, y, dparam, oprefix):
...
-    parser.add_argument("-oprefix", "--output_prefix", type=str, default="output", help="the prefix of the (various) output files.")
+    parser.add_argument("-A", "--action", choices=["conv", "deconv"], default=None, required=True, help="choose conv if you want to obtain the response of the thermostat on the vibrational density of states; choose deconv if you want to obtain the micro-canonical density of states by removing the disturbance induced by the thermostat")
+    parser.add_argument("-iFACF", "--input_facf", type=str, default=None, help="The input Fourier transform of the auto-correlation function (FACF). The unit of time is assumed to be atomic units unless the additional -time_scaling parameter is provided.")
+    parser.add_argument("-iXML", "--input_xml", type=str, default=None, help="The i-PI xml file that contains the Drift (A) and the Covariance (C) matrices as well as the ensemble temperature (T). The units are interpretted from the file and the A and C matrices are converted to atomic units. The C matrix is further divided by k_B T. If the -time_scaling parameter is provided, the units of A are modified using the scaling factor.")
```
ceriottm: Would be good to specify somewhere that it is assumed that the acf comes from the centroid, so in case of NMGLE, the centroid would be taken. Also, you could implement that if the input contains  pile_l, it treats it like a Langevin equation. Finally, if the input contains a thermo which is none of the above, an error should be thrown.

tools/py/gleacf.py
```diff
@@ -145,61 +145,83 @@ def ISRA(omega, ker, y, dparam, oprefix):
...
+            Cp = simul.syslist[0].motion.thermostat.C / kbT
+            Dp = np.dot(Ap, Cp) + np.dot(Cp, Ap.T)
+        if(ttype == "ThermoNMGLE"):
+            Ap = simul.syslist[0].motion.thermostat.A[0] * tscale
+            Cp = simul.syslist[0].motion.thermostat.C[0] / kbT
+            Dp = np.dot(Ap, Cp) + np.dot(Cp, Ap.T)
+        elif(ttype == "ThermoLangevin"):
+            Ap = np.asarray([1.0 / simul.syslist[0].motion.thermostat.tau]).reshape((1, 1)) * tscale 
+            Cp = np.asarray([1.0]).reshape((1, 1))
+            Dp = np.dot(Ap, Cp) + np.dot(Cp, Ap.T)
```
ceriottm: Add PILE-L, and an else: raise ValueError of some kind

tools/py/gleacf.py
venkatkapil24: Added a line on nmgle.

tools/py/gleacf.py
venkatkapil24: Added.

venkatkapil24:
I have changed them to small caps. However, I could not use -[singleletter] format for the sake of clarity. For instance there are many input options so instead of using i,j,k etc it is better to use ixml,ia, ic etc. Furthermore, there are many which begin with the same letter. getacf already uses small cap  command line options. Even if you agree with these PLEASE DO NOT MERGE YET! I want to make one more change to getacf.

ceriottm:
OK with me, I approve so you can merge as soon as you're done with the latest small tweaks

PR #14, opened 2019-01-26
Implementation of a temperature ramp.

ceriottm:
Minimal changes, just added a motion class to change the temperature. The depend mechanism made this a breeze. I'll merge straight away but leave the branch active to add perhaps a pressure ramp as well.


ceriottm:
Ended up adding a pressure ramp too. Really didn't need review (hopefully won't have bad surprises...)

PR #20, opened 2019-02-26
Improve documentation

mahrossi:
This only refers to very necessary improvements to the front matter of the documentation. There is no doubt (to me) that this should be merged ASAP. The doc is not bad, but the outdated information was making people give up on even looking at it.


ceriottm:
Looks very reasonable to me, but do you want to merge or to wait until we do more work?

mahrossi:
merge now, more work = further pull requests.

PR #22, opened 2019-03-05
Feat/clean outputs

ceriottm:
So, this is to try and clean up the mechanism by which different calculators print out auxiliary files that don't fit into the property/trajectory mechanism.
We should implement this with ALL of the motion classes that currently spit out stuff in random files. I started already with remd.


ceriottm:
@mahrossi the instanton example is a bit of a mess. there are scripts to run the different bits that assume paths in Yair's home folder. There's no readme that explains how to run  the example, etc. This could be a good opportunity to add those given you've already put some order in those folders.

mahrossi:
Seems we addressed all main issues.

PR #23, opened 2019-03-07
the ACF tool now also prints out the error (block avg) in the ACF.

venkatkapil24:
The code was already reading the time series in blocks. Implemented a simple error estimator.


venkatkapil24:
I agree completely. This kind of defeats the purpose.

ceriottm:
OK so go ahead and push something that accumulates the squares on the go, so you can get the variance at the end without storing every block. Afterwards I'll gladly merge.

PR #26, opened 2019-03-18
Feat/wrap unwrap

glensk:
This should be solely the wrap/unwrap changes.


tools/py/mux-positions.py
```diff
@@ -4,21 +4,61 @@
...
+    cell = frame['cell'].h
+    cell_inv = np.linalg.inv(cell)
+    s = cell_dot_pos(cell_inv,pos)
+    sc = s - np.round(s)
+    frame['atoms'].q = cell_dot_pos(cell,sc).flatten()
+    return frame
+
+
+def unwrap_positions(frame,framelast):
+    frame_wr = wrap_positions(frame)
```
ceriottm: Why do you wrap before unwrapping?

tools/py/mux-positions.py
```diff
@@ -53,8 +94,16 @@ def main(fns_in, fn_out, begin, end, stride):
...
+            for trj in trjs_in:                
                 frame = trj.next()
+
+                if wrap:
+                    frame = wrap_positions(frame)
+                if unwrap:
+                    frame = unwrap_positions(frame,frame_last)
+
+
+                frame_last = frame
```
ceriottm: There can be multiple trajectories, so you should store a frame_last for each trajectory. 
Something like holding a list sized as trjs_in, and traversing in together with trjs_in.

tools/py/mux-positions.py
glensk: did remove the wrap before unwrapping. Also, put wrap/unwrap to the bottom. Created separate functions for wrap unwrap due to your suggestion above: 
if wrap:
    wrap positions
if unwrap:
    unwraps positions

tools/py/mux-positions.py
glensk: done.

PR #32, opened 2019-03-30
Feat/planetary

mjwillatt:


ceriottm:
A few initial comments: 
1. Is it absolutely necessary to use scipy? Normally it's not one of i-pi's dependencies so pls see if you can use an alternative function that is implemented in numpy
2. Since running a planetary model simulation is not a trivial exercise, it'd be good to have a README in the example folder

PR #33, opened 2019-04-04
Fixed minor bug in socket.py

b-fg:
Minor typo bug in socket.py:  `raise socket.timeoout` to ` raise socket.timeout`


ceriottm:
Thanks for the (minor but appreciated) contribution! Keep these coming.

PR #35, opened 2019-06-15
Bufgfix/mts queue

ceriottm:
when using MTS with multiple force components on the same level, they get computed sequentially.
This creates a queue_mts function that should trigger all calculations at once.


mahrossi:
@fidanyan -- this should fix the troubles with force components. needs testing also for multiple stress tensors.

mahrossi:
Works with MTS and with SLC (and with MTS+SLC). Missing test for barostats still.

ceriottm:
Can this be merged? We have too many open PR

ceriottm:
@mahrossi can this be merged?

ceriottm:
Approved IRL by @mahrossi

PR #36, opened 2019-06-25
Geop dev

mahrossi:
This is a precocious pull request that is intended to revive this branch and fix some newly-encountered bugs. 

There have been reports of L-BFGS getting stuck for usual geometry optimizations (I have a test-case). The algorithm is used in the instanton tunneling splitting code and works there pretty well AFAIK, however it is called from another LBFGSOptimizer class that @litman90 wrote. 

@litman90 is on vacations this week, but should be back ASAP. I tried other optimizers in this branch and they seem just fine.


mahrossi:
P.S.: Please don't merge this branch too soon!

mahrossi:
Also reporting that the new default tolerance for the line search in https://github.com/i-pi/i-pi/pull/36/commits/d870ca73c31ba7a89f71a7e09f24f3ebfd950b31 fixes the problems with l-bfgs seen above. I don't see why the default should be 1 for the line search tolerance but if it was needed for any reason, let me know.

mahrossi:
Continuing on this topic, @fidanyan noticed that i-PI is back to doing one extra unecessary force evaluation per optimization step for all optimizers. @litman90 already had this problem more than 1 year ago and we implemented the `force_transfer` routine exactly for that. We need to check what went wrong again, something broke this.

PR #38, opened 2019-06-28
Geometry optimization with fixed atoms

mahrossi:
This is an implementation that makes a (much) better job at fixing atoms for geometry optimizations than what was there before. L-BFGS is not supported but the code stops if trying to do that at the moment. Needs some review to make sure the code is ok, but I have looked over it and it looks good.

Tests and examples are in examples/fhi-aims/h2o-geop-fixatoms (also includes an explicit comparison to ASE -- we can remove it).


mahrossi:
@litman90 can you review this too?

ceriottm:
Do you mean that l-bfgs is out, or that l-bfgs works but will stop if there
are fixed atoms?

fidanyan:
L-BFGS works as before if there are no fixed atoms, otherwise in throws a warning and stops.

litman90:
Hi,
I have tested trmbfgs fixing 1 atom and it **doesn't** work as expected.
The main problem is that the Hessian is not being updated which arises from the way the masking procedure is performed.

fidanyan:
Hello,
All algorithms work now, the only point to pay attention is that I changed default `ls_options["tolerance"]` from 1 to 1e-2, otherwise L-BFGS got stuck in `min_approx()`, thinking that the position is already converged. I think it's unrelated to fixatoms, even though I could not reproduce this behaviour in other cases. Any comments on `ls_options` are appreciated, also it is not necessary change since it can be provided in input file, so this part can be rejected without heavy consequences.

mahrossi:
I think you need even tighter criteria to converge there in many cases
(1e-4 or so). We had already discussed this long ago and I fixed this point
also in another branch already.

mahrossi:
@fidanyan what is the status here? have you thoroughly tested that _all_ algorithms work with fixed atoms like this?

fidanyan:
Yes, they passed all my tests. examples/lammps/h2o-geop-fixatoms/README and examples/lammps/h2o-geop-fixatoms/README contain the results.

examples/fhi-aims/h2o-geop-fixatoms/ase_comparison/log.ase
```diff
@@ -0,0 +1,1011 @@
+Accepting clients on INET port 31415
```
mahrossi: This file is probably not needed and will just inflate the repository. These type of files should never be committed.

examples/fhi-aims/h2o-geop-fixatoms/ase_comparison/opt.log
```diff
@@ -0,0 +1,40 @@
+      Step     Time          Energy         fmax
```
mahrossi: This and the next xyz files should be in a "regression test" type of folder. Not really in the examples folder. Examples are meant to tell people how to run something.

examples/lammps/h2o-geop-fixatoms/run-all-tests.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+. ~/soft/i-pi-mahrossi/env.sh
```
mahrossi: explicit paths in test scripts will obviously make them break for any other user.

mahrossi:
In addition: In the examples folder you can leave one or two examples (only with inputs) about how to run geop with fixed atoms and that should be enough.

fidanyan:
Done: I left only the minimal set in the examples/, and all the result files with an explanations are in tests/benchmarks/*.tgz.

fidanyan:
Moved test cases from tests/benchmarks/ to examples/regtest. `regtest.py` works for lammps, but for FHI-aims it doesn't work for some reason. Maybe later I'll try to sort it out. For now, I just left FHI-aims results as an archive with the explanations inside. It's quite clear.

PR #43, opened 2019-07-24
Cayley integrator for normal modes

korolrom:
I have added the Cayley integrator to normal_modes.py, trying to keep the same style.
The mass scaling (from sk) is also taken into account.

Implemented by taking A_i around eq. (25) to be [ [0, -s_k omega_k^2], [1/s_k, 0]] in
 https://arxiv.org/abs/1907.07941

This reproduces the exact integrator implemented in normalmodes.py if substituted in eq. (25) and the newly added Cayley transform if substituted in eq. (26)


mahrossi:
Hello there, just to clarify:

Did you substitute the usual normal-mode propagator by the Cayley propagator? That is what it looks like from the changes but I had only a fast look. If that is the case, could you do it in a way that you keep both options available? Perhaps as a separate function.

Then, could you add some better documentation for the propagator? Commenting lines or adding a doc string would be extremely helpful.

And lastly, could you also share possible tests you did, making sure that this works and does not break anything that was there before, etc.?

Thanks!!

korolrom:
Hi, Mariana!

This is my first edit of i-pi and I am not very familiar with the code structure, so I decided to do it carefully step-by-step.
I did not change either open paths or normal modes propagator, only added the Cayley version to this single file. I also did not change any control statements anywhere else.
Therefore, I don't expect anything to change when using the old version of the integrator.

However, the next step is to add an option to switch between the integrators and I wanted to double check before I do anything. In particular, should I just add a new option into the InputNormalModes and NormalModes objects?

In his e-mail Michele implied that I should implement it analogously to how the open paths integrator was added, but this one introduces open paths only for certain atoms. In the case of Cayley integrator presumably the option should be to either use Cayley or the exact propagator.

Thanks for your quick response!

Roman

ipi/engine/normalmodes.py
```diff
@@ -347,15 +347,15 @@ def get_prop_pq(self):
         # Note that the propagator uses mass-scaled momenta.
         for b in range(1, self.nbeads):
             sk = np.sqrt(self.nm_factor[b])
-            square = (self.omegak[b] * dt / 2)**2
-            pqk[b, 0, 0] = (1-square) / (1+square)
-            pqk[b, 1, 1] = (1-square) / (1+square)
-            pqk[b, 0, 1] = (4*square / dt * sk) / (1+square)
-            pqk[b, 1, 0] = dt / sk / (1+square)
-
+            dtomegak = self.omegak[b] * dt / sk
```
korolrom: Sorry, that was an accidental edit of the get_prop_pq(self) when I meant to only change get_cay_prop_pq(self).

I now reverted back to what it used to be before the edit

ceriottm:
Thanks for the contribution. I think the easiest thing at this point is really for you to implement get_cay_prop_pq and get_cay_oprop_pq to have monodromy matrices for closed and open path propagation. then I can guide you to implement an option (probably to normalmodes.py) to switch between the two propagators.

OndrejMarsalek:
When it comes to the input, I think it should go under `dynamics`, just like `splitting`. I realize that the code itself might be in `normalmodes.py`, but I think from a user's perspective, this is more natural.

ceriottm:
So, I suspect there is a mistake in how you wrote the propagator matrices, as this blows up in a second. See if you can fix it.

ceriottm:
I am merging, if you want to add anything you can open another PR. 
When the paper is out you may want to also make a PR on https://github.com/i-pi/web/ so you can document the new feature in http://ipi-code.org/about/features/

PR #46, opened 2019-08-06
It excludes the fixatoms from the hessian evaluation

litman90:
A small implementation that allows fixing atoms in phonon calculations. This is useful when only a block of the full hessian is needed.


litman90:
@mahrossi @ceriottm. 
It is a minor change. 
Just to be sure, I asked Karen to test it with his own completely independent implementation based on phonopy.
Everything looks ok.

fidanyan:
Hello, 
yes, I've tested several cases with my version of `Phonopy-FHI-aims` script (https://github.com/fidanyan/phonopy/blob/master/scripts/phonopy-FHI-aims). In most cases, they produce the same result. In the other cases, my script fails because it always applies additional symmetries, which are wrong because it doesn't see excluded atoms.

So, in the cases where I could compare, results are the same.

mahrossi:
@venkatkapil24 -- do you have anything against this implementation, as you have been the last person doing modifications to phonons? if not I will just immediately merge.

mahrossi:
I'll merge...

PR #49, opened 2019-08-30
RPC for open paths

litman90:
Use of the open_path variable when binds the Forces object. This allows having RPC with open paths.
If the paths are not all open or all close, the individual potential energy in the expanded RP are not well defined, even though its sum, the potential energy estimator and forces are.


ipi/utils/nmtransform.py
```diff
@@ -179,9 +179,13 @@ def b2nm(self, q):
 
         qnm = np.dot(self._b2nm, q)
         for io in self._open:  # does separately the transformation for the atom that are marked as open paths
-            qnm[:, 3 * io] = np.dot(self._b2o_nm, q[:, 3 * io])
-            qnm[:, 3 * io + 1] = np.dot(self._b2o_nm, q[:, 3 * io + 1])
-            qnm[:, 3 * io + 2] = np.dot(self._b2o_nm, q[:, 3 * io + 2])
+            try:
```
mahrossi: Does this introduce any time overhead in usual calculations?

ipi/utils/nmtransform.py
```diff
@@ -236,9 +244,12 @@ def b1tob2(self, q):
         """
         q_scal = np.dot(self._b1tob2, q)
         for io in self._open:  # does separately the transformation for the atom that are marked as open paths
-            q_scal[:, 3 * io] = np.dot(self._o_b1tob2, q[:, 3 * io])
-            q_scal[:, 3 * io + 1] = np.dot(self._o_b1tob2, q[:, 3 * io + 1])
-            q_scal[:, 3 * io + 2] = np.dot(self._o_b1tob2, q[:, 3 * io + 2])
+            try:
```
mahrossi: Same as above. Not sure there is any better way to do this, though.

mahrossi:
Hi, can we merge this now? @ceriottm can you confirm that although this is not the ultimate solution, the main issues you had are resolved?

ceriottm:
yes and yes.

PR #50, opened 2019-09-30
PIMD-B

BarakHirshberg:
Pull request to incorporate PIMD for bosons within i-pi. 
([link](https://arxiv.org/abs/1905.09053#) for all the details of the method on the arXiv)

**Main changes:**
1) In free_qstep() if propagator is "bab" a numerical velocity verlet step in Cartesian coordinates is done (possibly multiple time step with nmts being the number of internal steps). This is implemented in free_babstep().

2) For bosons, vspring and fspring are calculated in get_vspring_and_fspring_B(). This is done by calling Evaluate_VB() to get vspring and all the required energies for the forces). Then, forces are calculated in Evaluate_dVB() . In the end, a list containing [vspring, fspring] for bosons is returned.
(Along the way, Evaluate_EkN() and Evaluate_dEkN_on_atom() are called.)

3) Changed dself.fspring and dself.vspring to now hold the correct value for distinguishable particles, for bosons and for mixtures of both. This also makes sure that the converged quantity works in all cases.

4) Added regtests, /zundel-bab just to check to numerical propagator and /bosons with examples for bosons, comparison to distinguishable particles and a mixture of both.

5) Added 3D isotropic harmonic oscillator to driver.f90


ipi/inputs/normalmodes.py
```diff
@@ -87,8 +87,8 @@ class InputNormalModes(Input):
                                    "options": ['fft', 'matrix']}),
         "propagator": (InputAttribute, {"dtype": str,
                                    "default": "exact",
-                                   "help": "How to propagate the free ring polymer dynamics. Cayley transform is not exact but is strongly stable and avoid potential resonance issues.",
-                                   "options": ['exact', 'cayley']})
+                                   "help": "How to propagate the free ring polymer dynamics. Cayley transform is not exact but is strongly stable and avoid potential resonance issues. A bab scheme is also added to perform numerical verlet type propagation.",
```
mahrossi: Maybe add here that only bab works for bosons? (is that right?)

ipi/engine/normalmodes.py
```diff
@@ -94,8 +94,12 @@ def __init__(self, mode="rpmd", transform_method="fft", propagator="exact",
             freqs = []
         if open_paths is None:
             open_paths = []
+        if bosons is None:
+            bosons = []
         self.open_paths = np.asarray(open_paths, int)
+        self.bosons = np.asarray(bosons, int)
         dself = dd(self)
+        dself.nmts = depend_value(name='nmts', value=nmts)
```
mahrossi: I am just flagging this line to follow it up. Need to be a bit careful.

ipi/inputs/normalmodes.py
BarakHirshberg: Actually no, it works also for distinguishable particles, but is the only one that works for bosons.

mahrossi: Ok, then can you be more verbose in this respect in the places that this is called? Especially in the documentation?  I tried to submit a comment there too but it is not showing up here.

BarakHirshberg: OK, no problem. Please see the changes I just made below. Is that better? I think that these should be automatically included in the documentation but I'm not sure. If you want me to add more info anywhere else, could you point to the right place?

ipi/engine/normalmodes.py
```diff
@@ -18,7 +18,7 @@
 from ipi.utils import units
 from ipi.utils import nmtransform
 from ipi.utils.messages import verbosity, warning, info
-
+from ipi.utils.units import Constants #Perhaps not needed and beta can be found elsewhere?
```
venkatkapil24: Best to remove the line if redundant.

ipi/engine/normalmodes.py
```diff
@@ -499,7 +534,244 @@ def get_dynm3(self):
                     dm3[k, a] = self.beads.m3[k, a] * self.o_nm_factor[k]
         return dm3
 
+    def Evaluate_EkN(self, N, k):
```
venkatkapil24: I was wondering if it would be cleaner to create a separate file called ipi.utils.exchange.py (or a better sounding name) and import the functions from there ? May not be a bad idea since that could be more lines about fermions coming in the future.

ipi/engine/normalmodes.py
BarakHirshberg: Done, thanks for the suggestion.

ipi/engine/normalmodes.py
BarakHirshberg: Hi, I think that is a good idea. I tried moving everything to ipi.utils.exchange as you suggested. 
Can you have a look? Thanks.

ipi/engine/normalmodes.py
BarakHirshberg: Hi, not sure I understand what the problem is. Anything you need me to do?

mahrossi:
Merge with master branch passed -- let's make i-PI enter the era of bosons. Thank you @BarakHirshberg !

BarakHirshberg:
Great! Thank you guys @mahrossi and @venkatkapil24 !

PR #57, opened 2019-10-12
Feat/shutdown on disconnect

### LINKED ISSUE #55: i-pi & CP2K working on multiple beads
At the ICTP we aim to make a PRACE project proposal based on using i-pi together with CP2K, where i-pi computes path integrals among multiple beads running in parallel. The most efficient way it is normally to place one bead per each allocated node having i-pi running in background on one of the nodes. The main problem is that sometime i-pi excludes the nodes from the client lists. It seems like the problem can be solved by setting the timeout parameter of the .xml input file to either 0 or a big number, for example: 

[...]
    \<forces\>
       \<socket mode='inet' pbc='False'\>
          \<address\>r162c02s01\</address\> \<slots\>5\</slots\> \<port\>21153\</port\> \<latency\>10\</latency\> \<timeout\>5000\</timeout\>
       \</socket\>
    \</forces\>
[...]  

What I understand is that in both cases i-pi detects a client is not responding and it will continue to schedule the calculation on the remain nodes/clients. However, this is also inefficient as you would normally continue to run on the remain nodes while paying for all initially allocated resources. I guess the ideal would be that if some of the client doe not respond for a while the simulation terminates as a restart is needed at that stage. 

I wanted to ask if you confirm the way we are understating this and if you have any suggestions about how to end the simulation when a CP2K client is not longer (i.e., for several minutes) responding.  

Ivan Girotto (ICTP)

ceriottm:
Implements an option exit_on_disconnect for FFSocket that triggers a soft-exit event whenever a client disconnects or hangs.
Fixes #55


ipi/engine/properties.py
```diff
@@ -244,6 +244,9 @@ def __init__(self):
                           "help": "The value of the conserved energy quantity per bead.",
                           'func': (lambda: self.ensemble.econs / float(self.beads.nbeads))},
 
+            "ensemble_lp": {"dimension": "undefined",
+                                     "help": "The log of the ensemble probability",
+                                     "func": (lambda: self.ensemble.lpens)},
```
mahrossi: Why are these commits here?

ceriottm: Ehm.... probably cross-contamination from some other branch. These are harmless though so I'll just merge.

PR #58, opened 2019-10-14
Feat/classical constraints

ceriottm:
Implement classical constraints. 

TODO items

- [x] Clean up examples so that they all function, and it's clear what they do. 

- [x] If a folder contains several inputs include a README that explains what each input does

- [x] Document the code and the input. Include an explanation of the general algorithm, and refereences to the papers that discuss the theory of anything non-standard

- [x] Move the different constraints code in utils/constraints.py


ceriottm:
So this is getting really ready to merge. @venkatkapil24 can you start a quick but thorough code review? @GeorgeTrenins and @MatthiasSachs  can you add appropriate references to the comment lines?

venkatkapil24:
Sure. I will start having a look this afternoon.

examples/lammps/h2o-constraints/input.xml
```diff
@@ -1,56 +1,425 @@
...
+                    <maxit> 10000 </maxit>
+                    <norm_order> 2 </norm_order>
+                    <tolerance> .0001 </tolerance>
+                </csolver>
+                <nmts> [1, 4] </nmts>
+                <timestep units='femtosecond'> 4.0 </timestep>
+                <thermostat mode='langevin'>
+                    <tau units='femtosecond'> 5 </tau>
+                </thermostat>
+                <constraint mode="multi">
```
venkatkapil24: @ceriottm : Would it be possible to create some sort of a template based input -- as is done in REMD -- so that the redundancy can be removed ?

ipi/engine/motion/constrained_dynamics.py
```diff
@@ -0,0 +1,498 @@
+"""Contains the classes that deal with the different dynamics required in
```
venkatkapil24: Should this file exist ?

ipi/engine/motion/constrained_dynamics-old.py
```diff
@@ -0,0 +1,508 @@
+"""Contains the classes that deal with the different dynamics required in
```
venkatkapil24: Should this file exist ?

ipi/engine/motion/constrained_dynamics.py
```diff
@@ -0,0 +1,498 @@
...
+different types of ensembles.
+
+Holds the algorithms required for normal mode propagators, and the objects to
+do the constant temperature and pressure algorithms. Also calculates the
+appropriate conserved energy quantity for the ensemble of choice.
+"""
```
venkatkapil24: May be nice to update these as well.

venkatkapil24: Especially because the constraints have not been implemented with constant pressure dynamics. Do you think that bit can be quickly added ? Anyway I will return tomorrow morning.

ipi/engine/motion/constrained_dynamics.py
```diff
@@ -0,0 +1,498 @@
...
+        """ Solver options include a tolerance for the projection on the
+        manifold, maximum number of iterations for the projection, and 
+        the order of the norm to estimate the convergence """
+        
+        super(ConstraintSolver,self).__init__(constraint_list, dt)
+        self.tolerance = tolerance
+        self.maxit = maxit
+        self.norm_order = norm_order
+    
+    def proj_cotangent(self):
```
venkatkapil24: Would be nice to add some documentation here.

ipi/engine/motion/constrained_dynamics.py
```diff
@@ -0,0 +1,498 @@
...
+                x = np.linalg.solve(dstrip(constr.Gram), b)
+            else:
+                x = spla.cho_solve(dstrip(constr.GramChol), b)
+                
+            p[ic] -= np.dot(np.transpose(dg),x)
+        self.beads.p[0] = p
+        self.beads.p.resume()
+
+    def proj_manifold(self):
+
```
venkatkapil24: Would be nice to add some documentation here.

venkatkapil24:
A few things :

- [ ] docstring for functions like proj_manifold, proj_manifold
- [ ] design decision whether the program should work when there are no constraints or not. If not, the correct error message should be triggered.
- [ ] Calculation of effective number of degrees of freedom so that the simulations can give the correct temperature when small systems are studied.

ipi/engine/motion/constrained_dynamics-old.py
GeorgeTrenins: This is outdated and has now been removed.

venkatkapil24:
Is it already pep8 compliant ?

ceriottm:
> Is it already pep8 compliant ?

could you check?

PR #62, opened 2019-11-01
Add example instanton optimization with constrains

litman90:


venkatkapil24:
The example runs. How can I check if the result is correct ? This the output of inst.out :

    0.00000000e+00     6.50582761e-01   
    1.00000000e+00     6.46860332e-01   
    2.00000000e+00     6.46823708e-01   
    3.00000000e+00     6.46808552e-01   
    4.00000000e+00     6.46801597e-01   
    5.00000000e+00     6.46800793e-01   
    6.00000000e+00     6.46800508e-01   
    7.00000000e+00     6.46800467e-01

litman90:
The output is the expected one. You can also check that in the final instanton geometry only one atom is moving.

PR #65, opened 2019-11-06
Fixed atom geop regtests

mahrossi:
Revamp of regtest of geometry optimization with fixed atoms. Should be cleaner and work better now. Most changes by @fidanyan , I just cleaned the branch.


mahrossi:
We can erase the scripts now, Michele. The parent folder of the actual regtest doesn't rely on them in any way. We kept them there for ease of comparison in the future, but if you prefer them to be lost, sure. I would prefer the contrary but I will not impose it. I don't think they will be helpful in the tools folder.

ceriottm:
Do you think they'd be useful to someone wanting to see how to use the geop
module?I fear they may be confusing.
Overall, the examples/ folder has grown into a considerable mess. Yet
another "todo" item for i-pi v3!

mahrossi:
I think the regtests should just go out of the examples. There is also an example of geop with constraints.

venkatkapil24:
The geop regtests work on my PC. Let's make a quick decision on whether we want to keep these changes or not.

mahrossi:
I would keep these changes both in py2 and py3. Should we open a new pull request in py3?

examples/regtest/lammps/h2o-geop-fixatoms/analysis-scripts/012fixed.plt
```diff
@@ -1,7 +1,8 @@
 # Syntax on
 
 set term qt 0 position 100,200
-set key c b
+set key l t
```
mahrossi: all good.

mahrossi:
I'll merge this for "mirroring" regtests in py2 and py3.

PR #66, opened 2019-11-15
Cayley friction schedule condition

korolrom:
I have added a new method to compute the minimum tau_k as per condition (37) in https://arxiv.org/pdf/1907.07941.pdf. I need some guidance on how to proceed with the second condition. In particular, what are the frequency units in i-Pi and is there a way to access the highest frequency of the system as well as set it to some default value? That would be max_freq that is currently commented out.

Ondrej, I hope you can help with that.

Thanks, Roman

### DIFF
--- a/ipi/engine/thermostats.py
+++ b/ipi/engine/thermostats.py
@@ -345,6 +345,30 @@ class ThermoPILE_L(Thermostat):
 
         # Also include an optional scaling factor to reduce the intensity of NM thermostats
         return np.array([1.0 / (2 * self.pilescale * self.nm.dynomegak[k]) for k in range(1, len(self._thermos))])
+        
+    def min_tauk(self):
+        """Computes the minimum thermostat damping time scale for the non-centroid
+        normal modes given by the Cayley ergodicity condition (37) in https://arxiv.org/pdf/1907.07941.pdf
+
+        Returns:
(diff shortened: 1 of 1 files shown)


ceriottm:
Note that this should eventually be merged into master-py3 as master is now officially closed

ceriottm:
@korolrom  can you update this PR to refer to the master-py3 branch? Otherwise I will close it and you can create a new one against the Py3 version.

mahrossi:
Hello @korolrom -- this would be a good time to update the PR to the current master! Everything now is in py3 and this is a great implementation that should go there.

mahrossi:
I will merge this to this korolm-master here. Then we can try to open a PR to current master. At least it does not get lost, but korolrom-master branch should be deleted from this repo ASAP.

venkatkapil24:
What is the status here ?

korolrom:
Hi, sorry for prolonged silence. I just saw this.

Basically, we believe this modification is no longer needed as per our recent efforts to find the optimal friction for internal modes in the context of the BCOCB propagator (see https://arxiv.org/abs/2011.01601).

The condition was supposed to enforce underdamped dynamics, but we no longer think this is necessary (see ref. 52 of the paper https://arxiv.org/pdf/2011.01601.pdf).

I am happy to close this PR.

PR #70, opened 2019-11-26
Master py3

ceriottm:


ceriottm:
Just make master up-to-date before the apocalypse

PR #72, opened 2019-11-28
Update sGDML dependencies, README, and examples

sauceda:
Update sGDML dependencies, add some useful flags regarding the units, and parallelization of the predictor function.
Also, the ffsgdml example section was restructured and a README file was added. The previous models were updated since they were from an outdated sGDML version.


venkatkapil24:
Looks good. Also runs after installation of sGDML.

PR #73, opened 2020-01-08
Fixing tests in the py3 branch

ceriottm:
Many tests ended up breaking during the py3 porting process. 
The client interface seems to be particularly problematic. Help very much wanted.


mahrossi:
First, merge this pull request with the ASE client commented out. We can reinstate it at a later PR if necessary. Discussion ongoing regarding "native" ASE client.

PR #75, opened 2020-03-16
Adds fhi-aims regtest examples and corrects README in i-pi/test

litman90:


mahrossi:
Okay, I had a look. 

I believe that if I understand things correctly, `pars` is there to possibly send the client some parameters for initialization. It seems to be needed by QUIP and LennardJones forcefields if I followed it correctly. 

So, we can't disable `pars` completely and and pass a dummy msg instead to keep compatibility. Let us see if we can do something better, I'll try.

mahrossi:
Ok, this needs to be now finally tested and merged. String encoding is now uniform. If someone can attest that QE has no problems and that nothing is broken also in the forcefields that use parameters, i.e. LJ and QUIP, this is fine and should be merged ASAP!

PR #76, opened 2020-03-16
ASEClient

litman90:
This is an example that uses the ASE infrastructure to run the client code.
It shows how easy is to use all the ASE interfaces already created and tested (Gaussian, VASP, etc) with i-PI.

After creating an 'atoms' objects with ASE and attaching the corresponding calculator, one has to only add, for example:

from ase.calculators.socketio import SocketClient
port=10200
host='localhost'
client = SocketClient(host=host, port=port)
client.run(atoms)

I tested using the FHI-aims interface directly connected to i-PI obtaining the same results.
 
I also created a new folder so it is clear the distinction between the ASE interface inside i-PI and this example.


venkatkapil24:
The CP2K example doesn't work on my PC. Can @ceriottm or @mahrossi try as well ?

litman90:
For the records, On my computer, it works with the latest ASE-dev branch as well as the latest release that can be obtained from pip.

fidanyan:
I tried it with the latest CP2K 7.1 and it works.

PR #85, opened 2020-04-01
No fix for bug in pile_g plus fixatoms -- disabling functionality

mahrossi:
Addition of sanity check not to let users choose to fix atoms and fixcom True at the same time.


mahrossi:
Functionality will be disabled in this version.

PR #98, opened 2020-05-29
Black and flake8

litman90:
The code was reformatted using black and flake8 with some exceptions that need specific attention.
I'll report issues to address the exccetions after this PR is merged.
Note that the workflow now checks for formatting using flake8.


mahrossi:
I hope I solved all conflicts in the correct way -- it was a bit confusing in the kmc part.

mahrossi:
@ceriottm -- this is now ready to merge -- ok?

litman90:
Since this PR modifies several files I suggest creating a tag for the versions before and after this PR. 

It can be done by any user that has the right to push directly to this repo.
The commands are:

<on the current branch>
git tag -a v2.1.0 -m "Master-py3 before new formating"
git push origin v2.1.0

<After accepting the PR, and pulling to the local machine>
git tag -a v2.2.0 -m "Implementation of black+flake8 formatting"
git push origin v2.2.0

ceriottm:
Will try to merge this tomorrow - technically, today.

ceriottm:
OK finally getting to look at this PR. TODOs

- [ ] there should be a discussion in the documentation (ideally instructions in the manual, and referenced to in the README) that mentions which linting tools are used, and what is the coding convention

- [ ]  there should be a way to run the compliance tests locally (e.g. a Makefile with a "make lint" target)

- [ ] there should be a way to run the autoformatter so that one can quickly clean up code before pushing (e.g. a Makefile target "make pretty")

litman90:
I updated the README file with the corresponding instructions.
The application of black and flake8, which is explained in the readme takes care of point 2 and point 3.

litman90:
Ok, let me have a look at the black and flake8 docs to create some scripts.

ceriottm:
> Ok, let me have a look at the black and flake8 docs to create some scripts.

I really think a simple Makefile will do. You can also have a "test" target so people can do "make test". This is a pattern that is quite familiar to many people and so it'll be easier to make it part of the workflow of contributors

litman90:
I personally don't like adding extra layers when they are not necessary (either a Makefile or the script in python I've pushed). Particularly for these simple tools. 
However, I understand that in this way everything is "more normalized".
Since ipi is a python package, I guess it will be easier for any developer to modify the formatting script if it is written in python.

litman90:
Regarding the workflow, as far as I can tell about GitHub workflows, 
there is no much difference to have a Makefile or a python script.

ceriottm:
I'm not sure what is the problem with a Makefile, seems simpler and more
robust to me. it'd be a 4-liner

```
.PHONY: pretty, lint

lint:
       black
pretty:
       flake8
```

bin/i-pi-style
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python3
+
+"""
+This script cleans python files in-place by calling black.
+With option --path, it searches for python files
+recursively in the given directory. With option --file
+it applies autopep8 to listed python files.
```
mahrossi, line 7: this description is wrong.

litman90:
> Thanks Yair for the script, looks good. I am not sure I agree with having one more layer too -- the last one we had broke soon enough. Anyways this one is simpler.
> 
> It will anyways be a job of the people who accept merge requests (us), to ensure and request that they follow these steps.

litman90:
> Thanks Yair for the script, looks good. I am not sure I agree with having one more layer too -- the last one we had broke soon enough. Anyways this one is simpler.
> 
> It will anyways be a job of the people who accept merge requests (us), to ensure and request that they follow these steps.

Please note that the workflow will fail if the format of the files is not appropriate. In this sense, the formatting is implicitly enforced.

It is possible to enforce it explicitly by adding the formatting steps to the workflow. I don't think it is a good idea to do it though .

mahrossi:
Yes, our job will still be to request that they do it and never merge until everything passes.

PR #100, opened 2020-06-07
Option to read wildcards for replay of many beads

fidanyan:
It is possible now to use globs in a trajectory filename in `replay` mode. 
If a filename with glob is provided, i-pi finds all files in working directory that match the provided pattern (Unix globs *?[] ), reads bead indices from them, assuming that these files have i-pi output format (XYZ or PDB). Then, it assigns the coordinates from these files to the corresponding beads. 
Consistency of input is checked, only 'conservative' use (beads from 0 to Nbeads-1) is allowed.


fidanyan:
I'll make a PR to `i-pi-mahrossi` instead

ceriottm:
This needs to be rebased after we merge the flake8 branch

PR #103, opened 2020-06-17
Examples + reg tests + update instanton

litman90:
Two new groups of tests are included within pytest:

ipi_tests/examples/test_examples.py:

This test goes through all the examples in the example folder, tries to run 2 steps of the simulation, and checks that i-pi closes without errors.
The list inside ipi_tests/examples/excluded_test.txt excludes some examples.
I still have to check several of them. Afterward, I will open issues to fix each of them.

ipi_tests/regression

Some examples of regression tests where the simulation output is checked against a reference output.

The instanton updates include a new feature to not evaluate forces in all bead but used an energy and/or distance criteria instead.


examples/MBPOL/splitting/TS/input.xml
```diff
@@ -5,7 +5,7 @@
         <trajectory stride='1' filename='pos'>  positions{angstrom}  </trajectory>
    </output>
    <total_steps>400       </total_steps>
-   <ffsocket name="aims" mode="unix" >
+   <ffsocket name="qtip4pf" mode="unix" >
```
mahrossi, line 8: Why is this inside "MBPOL" splitting folder and apparently uses as a driver qtip4p?

mahrossi: P.S.: Other files in the folder are the same.

examples/lammps/isof-water/input.xml
```diff
@@ -8,7 +8,7 @@
   <prng>
     <seed>32345</seed>
   </prng>
-  <ffsocket name='lmpserial' mode='unix'>
+  <ffsocket name='qtip4pf' mode='unix' pbc='false'>
```
mahrossi, line 11: is pbc=false needed here?

examples/pes/instanton/ch4h-cbe-rate/README
```diff
@@ -3,7 +3,7 @@ TODO: Update with detailed instructions
 Folders
 TS: Finds classical transition state 
 reactant: minimizes and calculates phonons for the reactant, necessary for the final calculation of the rate
-instanton: finds instanton geometries with 80 and 40 beads at 300 K.
+instanton: finds instanton geometries using different implement features
```
mahrossi, line 6: *implemented

examples/MBPOL/splitting/TS/input.xml
litman90: The automatic tester looks for the name variable in ffsocket to assign a driver model.
If it recognizes it uses the corresponding model. Otherwise uses gas.

These are old examples that I used do check the code against JOR results using MBPOL.

examples/lammps/isof-water/input.xml
litman90: This tag controls if ipi applies pbc before passing the data to the client.
Since qtip4pf built-on driver is not pbc proof, it is safer to set it to false.

examples/MBPOL/splitting/TS/input.xml
mahrossi: We need to discuss this infrastructure.

examples/lammps/isof-water/input.xml
mahrossi: Yes, I am fully aware about what the flag does. But now I see that again here we have this naming problem (you are calling qtip4p from the driver, but the example is in lammps folder). Let's try to find a better solution for this, this will become very confusing to future users.

mahrossi:
After (live) conversation last Thursday: The infrastructure will be slightly updated to better handle the issues above and a new PR will be issued. Many THANKS @litman90 !!!

litman90:
This PR is going to be divided is several pieces.
See PR #105 and #106.

PR #105, opened 2020-06-22
Only examples

litman90:
This PR comes after an offline discussion I had with MR.

As the title says, only example testing is updated.

ipi_tests/examples/test_examples.py:

This test goes through all the examples in the example folder, tries to run 2 steps of the simulation, and checks that i-pi closes without errors.

The list inside ipi_tests/examples/excluded_test.txt excludes some examples.

I still have to check several of them. Afterwards, I will open issues to fix each of them.
The script is quite flexible and can be run locally to test all the examples, a subfolder, or just one.


ipi_tests/examples/exampletools.py
```diff
@@ -0,0 +1,186 @@
...
+              Error {}".format(
+                    str(cwd), ipi.communicate(timeout=2)[0]
+                )
+            )
+
+        except FileNotFoundError:
+            raise ("{}".format(str(cwd)))
+
+        except ValueError:
+            raise ("{}".format(str(cwd)))
```
mahrossi, line 186: I would like to have a quick Slack chat about this script in the common channel.

ipi_tests/examples/modify_xml.py
```diff
@@ -0,0 +1,44 @@
...
+        new_element = ET.SubElement(root, "total_steps")
+        new_element.text = str(nsteps)
+
+    tree.write(open(output_name, "wb"))
+
+
+if __name__ == "__main__":
+    name = sys.argv[1]
+    out = "output.xml"
+    modify_xml(name, out)
```
mahrossi, line 44: Why both this script and the one above (examplestools) are needed?

litman90: I am not sure if understand the question. I try an answer though.

The section highlighted above allows calling the function modify_xml for debugging
or experimenting purposes.
The exampletools file contains the object definition and functions required
to find all the examples, to check if they are in the excluded list or not, to create the temporary folders, to modify the xml file (change the number of steps to 2, etc)  and run the examples.

litman90: The address names are changed in such a way that if one test fails, and the ipi_localhost is not erased, the other tests can still run.

ipi_tests/examples/exampletools.py
litman90: Sure

ipi_tests/examples/modify_xml.py
litman90: In principle, it is possible to create a modify_xml function that can deal with all the necessities of examples and regression tests.
However, it is not desirable to have two simultaneous PR that modify the same file. The reason is that unless they are merged without further modification, which in general should not be assumed, conflicts will appear.

One way to proceed is that after the PRs are accepted, one can create an issue to discuss if it is better to entangle the two tasks ( checking examples and regression tests) or it is better to let them evolve independently.

I actually would say that, even though in general cases code repetition is a bad practice, for this particular situation it is better to leave them unentangled to have 2 simple functions instead of having a big one that considers all the possibilities and corner cases and will be more difficult to maintain and modify.

venkatkapil24: I also agree that it is a bit confusing to have two modify_xml files. Perhaps the two functions should have different names.

ipi_tests/examples/test_examples.py
```diff
@@ -0,0 +1,92 @@
...
+
+examples_folder = Path(__file__).resolve().parents[2] / "examples"
+excluded_file = Path(__file__).resolve().parent / "excluded_test.txt"
+
+examples = find_examples(examples_folder, excluded_file)
+print("We have found {} reg_tests".format(len(examples)))
+
+
+@pytest.mark.parametrize("ex", examples)
+def test_example(ex):
```
venkatkapil24, line 21: Docstring

ipi_tests/examples/modify_xml.py
mahrossi: I think different names is good for now too.

ipi_tests/examples/test_examples.py
litman90: ok

ipi_tests/examples/modify_xml.py
litman90: It seems that my idea of having a different file for testing and experimenting purposes is creating more confusion than what it may help future developers. For now, I think it's better to delete it.

PR #106, opened 2020-06-22
Only regression test

litman90:
This PR comes after an offline discussion I had with MR.

As the title says, only the regression tests are updated.

ipi_tests/regression

Some examples of regression tests where the simulation output is checked against a reference output.


ipi_tests/regression_tests/slc/input.xml
```diff
@@ -0,0 +1,50 @@
+<simulation verbosity="high">
+  <ffsocket mode='unix' name='wat1beads' pbc='true'>
+    <port> 33330 </port>
```
mahrossi, line 3: is there any reason to add a port to a unix socket here? leaving it in tends to create confusion to users sometimes, even though it does not harm in any way. maybe a good option would be using inet/localhost also here just to make it clear how it would work with real different ports? not super important.

litman90: are you suggesting changing everything to inet/localhost? or to add at least one regression test where inet/localhost is used?

mahrossi: I think there are already tests where inet comm is used. In slc it is useful just to show the setup using different ports. So yes, I would tend to try inet/localhost in this case. But this should not be a reason not to merge this ASAP. @venkatkapil24 and @ceriottm -- fine?

ipi_tests/reg_test/runstools.py
```diff
@@ -1,46 +0,0 @@
-import subprocess as sp
-from pathlib import Path
-import pytest
-import time
-import os
-import shutil
-import tempfile
-
-
-class Runner(object):
```
venkatkapil24, line 10: Can you please add the description of the functions using docstrings ?

ipi_tests/reg_test/runstools.py
```diff
@@ -1,46 +0,0 @@
...
-    def __init__(self, parent):
-        self.parent = parent
-
-    def _run(self, cmd1, cmd2, cwd):
```
venkatkapil24, line 14: Same for all the functions that you have defined. Can you add some description using docstrings ?

ipi_tests/reg_test/runstools.py
```diff
@@ -1,46 +0,0 @@
...
-        except AssertionError:
-            raise AssertionError("{}".format(str(cwd)))
-
-        except FileNotFoundError:
-            raise FileNotFoundError("{}".format(str(cwd)))
-
-        except ValueError:
-            raise ValueError("{}".format(str(cwd)))
-
-    def _check_error(self, ipi, driver):
```
venkatkapil24, line 45: Here as well.

ipi_tests/reg_test/test_geop.py
```diff
@@ -1,33 +0,0 @@
...
-        "i-pi-driver -h localhost -p 33334 -m ch4hcbe",
-        "geop/bfgs",
-        "min.out",
-    ],
-    ["i-pi input.xml ", "i-pi-driver -u -m ch4hcbe", "geop/sd", "min.out"],
-]
-
-
-@pytest.mark.parametrize("cmd1,cmd2,folder,file", cmd1_cmd2_folder_output)
-def test_cmd_and_files(cmd1, cmd2, folder, file):
```
venkatkapil24, line 24: Can you add description here as well ?

ipi_tests/regression_tests/runstools.py
```diff
@@ -0,0 +1,114 @@
...
+import numpy as np
+import time
+import glob
+import os
+import shutil
+from tempfile import TemporaryDirectory
+import tempfile
+
+
+def get_info_test(parent):
```
venkatkapil24, line 14: Comments are missing in this file. Again it becomes tougher to follow without docstrings.

ipi_tests/regression_tests/runstools.py
```diff
@@ -0,0 +1,114 @@
...
+class Runner(object):
+    def __init__(self, parent, check_errors=True, check_prop=True):
+        self.parent = parent
+        self.check_error = check_errors
+        self.check_prop = check_prop
+
+    def _run(self, cmd1, cmd2, cwd):
+
+        try:
+            for filename in glob.glob("/tmp/ipi_*"):
```
venkatkapil24, line 56: Note that if another test is running it may fail to produce a RESTART file in case the /tmp/Ipi_$WHATEVER file is absent.

ipi_tests/regression_tests/runstools.py
```diff
@@ -0,0 +1,114 @@
...
+
+        except ValueError:
+            raise ValueError("{}".format(str(cwd)))
+
+    def _check_error(self, ipi):
+        ipi_error = ipi.communicate(timeout=30)[1].decode("ascii")
+        print(ipi_error)
+        assert "" == ipi_error
+
+    def _check_properties(self, cwd):
```
venkatkapil24, line 99: What about other properties ? Currently it looks like it only checks for a similaton.out file. What if there are more than one property files ?

ipi_tests/regression_tests/geop/bfgs/input.xml
```diff
@@ -1,10 +1,10 @@
...
-    <output prefix='min'>
+    <output prefix='simulation'>
         <properties stride='1' filename='out'>  [ step, potential{electronvolt}] </properties>
         <trajectory stride="1" filename="xc" format="xyz">x_centroid{angstrom}</trajectory>
    </output>
    <total_steps>400       </total_steps>
-   <ffsocket name="cbe" mode="inet" >
+   <ffsocket name="cbe" mode="unix" >
        <port> 33334 </port> 
        <address> localhost </address>
```
venkatkapil24, line 9: This is rather confusing. Perhaps should use another address for unix socket.

ipi_tests/regression_tests/runstools.py
mahrossi: Here I agree, but I think other properties can be added later on. This is supposed to be a minimal "skeleton" of the infrastructure, so definitely adding a doc string with a hint as to how one could enhance this function (other properties) would be a good way to go.

ipi_tests/regression_tests/slc/input.xml
litman90: No problem, I'll change this example to use inet/localhost

ipi_tests/reg_test/test_geop.py
litman90: This file was deleted

ipi_tests/regression_tests/runstools.py
litman90: The tests run sequentially so this will not generate problems.
Of course, if someone is running i-pi and wants to do a regression test on the same computer, this may be inconvenient.
But,
1. This won't happen in the GitHub workflow
2. I find it much more probable that someone will try to run regression tests several times because something is failing and this function helps that task.
3. Last but not least, in the future, we will have hopefully many regression tests, so if some of them fail this function
cleans the stage for the remainder tests

ipi_tests/regression_tests/geop/bfgs/input.xml
litman90: I am sorry, I don't get what is confusing. Could you be more specific?

ipi_tests/regression_tests/runstools.py
litman90: I added some dummy functions. I have some ideas but it would be better to discuss them first.

PR #107, opened 2020-07-06
Add python client class back

tomspur:
The python client class was deleted in a recent commit 3f3a42608 when the ASEClient was removed. The client class is required to add the i-pi interface to psi, see https://github.com/psi4/psi4/pull/1057

Can the Client class, please stay part of the code, so that the connection to psi works as well?

### DIFF
--- a/ipi/interfaces/__init__.py
+++ b/ipi/interfaces/__init__.py
@@ -5,4 +5,4 @@
 # See the "licenses" directory for full license information.
 
 
-__all__ = ["sockets"]
+__all__ = ["sockets", "clients"]
--- a/ipi/interfaces/clients.py
+++ b/ipi/interfaces/clients.py
@@ -147,10 +147,10 @@ class Client(DriverSocket):
                     i_step += 1
                 elif msg == Message("getforce"):
                     self.sendall(Message("forceready"))
-                    self.sendall(self._potential, 8)
+                    self.sendall(self._potential)
                     self.sendall(self._nat, 4)
-                    self.sendall(self._force, 8 * self._force.size)
-                    self.sendall(self._vir, 9 * 8)
+                    self.sendall(self._force)
+                    self.sendall(self._vir)
                     self.sendall(np.int32(0), 4)
                     self.havedata = False
                 else:
--- a/ipi/interfaces/clients.py
+++ b/ipi/interfaces/clients.py
@@ -1,0 +1,236 @@
+"""Clients that get interactions from specific programs.
+
+Here we implement classes that provide interactions obtained from specific
+programs and serve them back to i-PI.
+"""
+
+# This file is part of i-PI.
+# i-PI Copyright (C) 2014-2015 i-PI developers
+# See the "licenses" directory for full license information.
+
+
+import sys
+import os
+import socket
+import time
+
+import numpy as np
+
+from .sockets import DriverSocket, Message
+from ..utils import units
+
+
+class Client(DriverSocket):
+
+    """Base class for the implementation of a client in Python.
+
+    Handles sending and receiving data from the client code.
+
+    Attributes:
+        havedata: Boolean giving whether the client calculated the forces.
+    """
+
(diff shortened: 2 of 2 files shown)


litman90:
Since last year ASE supports socket communication using the ipi protocol.
It was decided to not support that in-built ASE interface anymore since in
principle it is not needed.
I suggest you have a look at i-pi
<https://github.com/i-pi/i-pi/tree/master-py3>/examples
<https://github.com/i-pi/i-pi/tree/master-py3/examples>/*ASEClient*/ in the
master-py3 branch, where you will find examples of how to run i-pi with
ASE.
Note that there are two ways of doing that.
Does it solve your problem?

ceriottm:
Hi @tomspur - sorry if we broke your workflow. The rationale was that ASEClient was duplicating a feature that had been available on the ASE side for years, so we thought to take it out to simplify the code. Would it be a solution for you to bring in the client code on the psi side?

tomspur:
No worries, changes happen :) I am fine with dropping the ASEClient and only would require the Client class.
My rationale for keeping it here is that the Client and Driver can more easily be kept in sync, e.g. when the communication may change in the future in the client code (https://github.com/i-pi/i-pi/blob/master/ipi/interfaces/clients.py#L150) or the driver code (https://github.com/i-pi/i-pi/blob/master/ipi/interfaces/sockets.py#L375). E.g. there are other ways to communicate it between the client besides the current inet/unix or other informations besides the current positions/forces, etc might be passed on in the future.
If those classes are in multiple different locations (e.g. here, ASE, psi, ....) it is more difficult to remain compatible among the versions and I would subclass the Client class in psi (similarly, like I expect ASE to do the same. A comment about such compatibility can also be seen e.g. here https://github.com/i-pi/i-pi/blob/master/ipi/interfaces/sockets.py#L387 :)

If you prefer me to push the client code to the psi side instead to simplify the code above, let me know and I'll do so

ceriottm:
I kind of agree that having a Client interface is not a bad idea, also to facilitate external codes which want to include it (e.g. this very case!) @litman90 you were the one who found issues with the ASEClient functionality - if I recall the problem was really in the ASEClient so I see no harm if @tomspur brings back (and tests thoroughlt ^_^) the Client part.

litman90:
Hi,
I see no problem with bringing it back as far as we ensure that it is
bug-free or at least delivers the exact same results as the inet/unix
interface.

mahrossi:
Hello @tomspur -- are there examples you tested this with?

mahrossi:
Hey @tomspur -- any news? If you can provide some examples we can test here independently and see if we can merge this?

tomspur:
Hi @mahrossi 

one example is the one from qespresso in `examples/qespresso/h2o/input.xml` from [here](https://github.com/i-pi/i-pi/tree/master/examples/qespresso/h2o)

and from the psi side, this input works to get the driver running with the H2O there as well
```
water = psi4.geometry("""
  O -1.216  -0.015  -0.261
  H -1.946   0.681  -0.378
  H -1.332  -0.754   0.283
  units angstrom
  no_reorient
  no_com
""")

psi4.set_options({
    'scf_type':             'df',
    'basis':                'cc-pvdz',
    'freeze_core':          'true',
    'dft_radial_points':    99,
    'dft_spherical_points': 302,
    'e_convergence':        8,
    'd_convergence':        8
})

options = {}

psi4.ipi_broker("bp86-d2", serverdata="inet:127.0.0.1:3141", options=options)
´´´

but also many other calculations are possible, when passed to the ipi_broker there.

Are you able to test that as well?

mahrossi:
Hey @tomspur -- thanks. 

In the meantime since my last comment, we have made the master branch py3 compliant (basically, master-py3 became master). Could you re-open this PR on master, add an example including a README on for a the calculation above for Psi with i-PI  and make sure that the current example in the current master branch works? 

Note that now we enforce black/flake8 formatting with the CI. Check out the README in the main repo for some details. 

Then the functionality will have less probabilities to get broken in the future ;-) and then I will also test it here. Many thanks for the contribution!!!

Issue #108, opened 2020-07-07
Phonons with ASR: problem in python3 branch.


arpank-2380:
Hello everybody,

I have encountered one issue with the python-3 branch for calculating phonon frequencies. I ran the example with water. It seems that FD, NMFD and ENMFD code is breaking while applying the acoustic sum rule and gives me the following error in the ipi output. Although it is doing all the displacements but not calculating the dynamical matrix, eigenvectors, etc. This is not happening with the officially released i-PI with python-2. If I do not use <asr> attribute, then it is working fine.
`Hello everybody,

I have encountered one issue with the python-3 branch for calculating phonon frequencies. I ran the example with water. It seems that FD, NMFD and ENMFD code is breaking while applying the acoustic sum rule and gives me the following error in the ipi output. Although it is doing all the displacements but not calculating the dynamical matrix, eigenvectors, etc. This is not happening with the officially released i-PI with python-2. If I do not use <asr> attribute, then it is working fine.

Exception in thread Thread-10:
Traceback (most recent call last):
  File "/software/python-3.5.2-el7-x86_64/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/software/python-3.5.2-el7-x86_64/lib/python3.5/threading.py", line 862, in run
    self._target(*self._args, **self._kwargs)
  File "/home/arpank/i-pi-master-py3/ipi/engine/motion/phonons.py", line 114, in step
    self.refdynmatrix = self.apply_asr(self.refdynmatrix.copy())
  File "/home/arpank/i-pi-master-py3/ipi/engine/motion/phonons.py", line 263, in apply_asr
    R[iatom, 1] * U[idof, 2] - R[iatom, 2] * U[idof, 1]
IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices

Does anyone have any ideas why applying asr for the python3 branch is creating a problem? 
I have attached a tar file containing my tests with py2 and py3 version of the code. `

Does anyone have any ideas why applying asr for the python3 branch is creating a problem? 
I have attached a zip
[h2o-lammps-test.zip](https://github.com/i-pi/i-pi/files/4886610/h2o-lammps-test.zip)
 file containing my tests with py2 and py3 version of the code.

ceriottm:
Try to change line 260 to i//3. It's the dreaded integer division problem
when converting py2 to py3.
If this work we'll run a quick bugfix PR.
[...]

PR #109, opened 2020-07-08
The integer division strikes back

ceriottm:
stray / 3 in the acoustic sum rule functionality of the phonons calculator


ceriottm:
Fixes #108

PR #111, opened 2020-07-08
Uncommented lines that enable constant pressure SC PIMD

venkatkapil24:
For some reason the lines which define stress for scpimd were commented out :|


ceriottm:
Does this explain the pathetic performance of NpT in Giulio's calculations?

PR #112, opened 2020-07-09
Fix again on the actual py3 branch....

ceriottm:


ceriottm:
oops! I had actually fixed this to the py2 master branch.

PR #115, opened 2020-07-21
Feat/example vasp

venkatkapil24:
Examples that work with VASP 6.1


venkatkapil24:
It was me just being silly in setting up the i-PI input. The interface seems solid.

PR #118, opened 2020-07-23
Only instanton3.0

litman90:
This PR is the last part of the PR #103 which was divided into three PR.
Here, the instanton functionality is updated to include a new feature that allows to not evaluate forces in all bead but use
an energy and/or distance criteria.

Examples and regression tests are included.


litman90:
I close it there to do a PR to master

PR #122, opened 2020-07-28
  Deleted "garbage" files that got erroneously merged in the past.

mahrossi:
Discovered some files that should not have made into the repository. I think they came from @litman90 (according to blame ;) ). I deleted them. Hope it is a smooth merge.


examples/dev_tests/lammps/h2o-geop-fixatoms/bfgstrm-0fixed/input.xml~HEAD
```diff
@@ -0,0 +1,38 @@
+<!-- REGTEST
```
mahrossi, line 1: I will fix this before merging.

mahrossi:
We do _not_ understand why this is still trying to run circleCI instead of our actions. This is very strange!!

mahrossi:
According to myself this is now ready to be merged and should/could be merged without much reviewing (pretty basic restructuring of non-essential files, plus making sure that the CI works properly)

litman90:
Everything looks fine.

PR #127, opened 2020-08-17
Examples repair

eszterpos:
This pull request fixes some non-working examples.
The exampletools.py was only copying files, but not subdirectories from the example directory upon running, now it's copying subdirectories, too. 
The lj driver was causing a problem in some cases, where additional flags "-o" were missing, an option for extra flags is now available and should be read in from the .driver_info file
The dev_tests and pes_regtest moved out from the examples/


ipi_tests/examples/exampletools.py
```diff
@@ -30,6 +31,9 @@ def get_driver_info(
     socket_mode="unix",
     port_number=33333,
     address_name="localhost",
+    flags1=0.0,
```
litman90, line 34: It would be better to make this more general, i.e. not enforce a fixed number of flags.

ipi_tests/examples/exampletools.py
```diff
@@ -59,6 +67,9 @@ def get_driver_info(
         "socket_mode": socket_mode,
         "port_number": port_number,
         "address_name": address_name,
+        "LJ_flags1": flags1,
```
litman90, line 70: It would be better to make this more general, i.e. not only for LJ potential

ipi_tests/examples/exampletools.py
```diff
@@ -84,15 +98,21 @@ def find_examples(parent, excluded_file="excluded_test.txt"):
         if os.path.isfile(Path(ff) / "input.xml"):
             if ff not in excluded:
                 examples.append(ff)
+                print(u"\u2713", " ", ff)
```
litman90, line 101: Why is this necessary?

ipi_tests/examples/exampletools.py
```diff
@@ -108,10 +128,13 @@ def modify_xml_2_dummy_test(
                 dd = driver_info["address_name"] + "_" + str(nid) + "_" + str(s)
                 element.text = dd
                 address = dd
+            flags1 = driver_info["LJ_flags1"]
```
litman90, line 131: Same as above.

ipi_tests/examples/excluded_test.txt
```diff
@@ -1,46 +1,22 @@
...
-hswfqmc/FixWF_1Bead 
-hswfqmc/WFOpt_2Bead 
+MBPOL/splitting/instanton/32-lbfgs                                          #This is resolved with PR #123  
+pes/qtip4pf/tun-splitting/instanton/32-lbfgs                                #This is resolved with PR #123  
+lammps/h2o-planetary-64                                                     # TypeError: string argument expected, got 'bytes' 
 
 # Broken
 ffdebye/single_particle 
 lammps/h2o-constraints 
+lammps/isof-vapor                                                           # Isotope-zeta-sc is stopping the termination of the run
```
litman90, line 9: Please after this PR gets merged, generate issues corresponding to these examples.

ipi_tests/examples/exampletools.py
eszterpos: I removed this, it wasn't necesssary, I just had it there for debugging checks.

eszterpos:
I added the option for reading in multiple flags with flexible number of variables from different lines of the .driver_info file.

I also added an option to flexibly change the number of manually specified folders to run the examples from. 
          python ipi_tests/examples/test_examples.py -f folder1 folder2 ...

ipi_tests/examples/exampletools.py
eszterpos: see general comment below

ipi_tests/examples/exampletools.py
eszterpos: see general comment below

ipi_tests/examples/exampletools.py
eszterpos: see general comment below

ipi_tests/examples/excluded_test.txt
eszterpos: yes, will do!

litman90 (approved):

litman90:
This PR extends a bit the infrastructure to test the examples provided with the code. 
Additionally, it fixed some tests which were excluded before. 
Since it does not touch any core functionallity, I consider that no further review is required and procced to merge it in master.

PR #129, opened 2020-09-11
Adding stop if stress tensor is completely zero.

mahrossi:
Presumably if someone really wants to do some simulation like this
  they can disable this.

  As a note, also found that the ValueError call in
  my computer repeats itself "infinitely" in the loop, apparently
  this is related to the threads. Could not find an immediate solutions.
  Anyone knows? It affects all ValueError calls in i-PI in the py3
  version, it seems. At least in my mac.


mahrossi:
P.S.: Tested on the FHI-aims example for water-npt in the examples folder of i-PI.

mahrossi:
Oh hmm: @litman90 -- this check and stop breaks a few of the small automatic tests of the CI because they were relying on i-PI just accepting to run a bit even without a virial... can we think of a way of still maintaining the CI but leaving, of course, this stop call in?

ceriottm:
Actually, I think this should only break if it's running a npt simulation.
Computing the stress in plane wave codes is not completely cheap so there
is a use case for not doing it in a NVT run.

mahrossi:
Yes, but it breaks only for NPT and NST indeed. Not NVT.

litman90:
I see three possible solutions. Naturally, there is a compromise between
the work needed now and the quality of the solution:

- to create/extend/modify a driver that can deliver some non-zero stress
tensors, at least of this dummy testing purposes
- to figure out how to install lammps in the repo and only provide NPT
examples which use lammps
- to add to the "excluded examples" list the NPT simulations until the
previous point is fulfilled

mahrossi:
I think that not testing the NPT examples is a bad idea even if only for a short time (because in practice, how short it will be is unclear). Going for the first option is the best and maybe we already have a dummy example that can do it. We try.

ceriottm:
I think that the para-h2 and LJ drivers also compute the stress tensor.

mahrossi:
harm3d is the simplest and fastest which already works for these tests. I started doing the modifications as you see but the option still needs to be passed and I ran out of time for today now. I did not have time to read and try the scripts to see where to pass them exactly.

litman90 (commented):
I am preparing another PR to make it possible to provide some arguments specific to the driver, which is needed here.

mahrossi:
Thanks Yair, but as far as I can see after looking a bit more, if I added:
"flags -o 2" to all these .driver_info it would already work. Am I missing something?

litman90 (approved):

litman90:
Indeed the possibility of reading the drivers' flag was already implemented, I pushed the modifications that fix the pipeline.

litman90:
The modifications proposed in this PR are only safety measures and all the tests pass without error.
I proceed to merge it.

PR #130, opened 2020-09-14
Regtest infrastructure

eszterpos:
The following pull request is to set up a basic regression test infrastructure for i-pi.
We create functions that are checking if the defined output files are 'all-close' to the output files verified by our independent implementations. These should be listed in the files_to_check.txt file in each new regtest directory.

Currently available file formats for comparison: .xyz and numpy accessible files.

I also deactivated the NEB routine until further bugfixes are conducted as its results could not be reproduced by independent tests.


litman90:
Let first make the test pass and resolve the conflicts.

PR #132, opened 2020-09-24
Regtest infrastructure

eszterpos:
The following pull request extends the existing regression test infrastructure for i-pi and redefines some of its basic features.

The github pytest environment is now changed to be based on the conda repository instead of pip. This enables proper numerical comparison between reference and the test run.

We define functions that check if the defined output files are 'all-close' to the output files verified by our independent implementations. These should be listed in the files_to_check.txt file in each new regtest directory.

Currently available file formats for comparison: .xyz and numpy accessible files.

Currently included regression tested i-pi features are:
   - NVE dynamics single and multi-bead
   - NVT dynamics with several thermostats single and multi-bead
   - Geometry optimzation routines: BFGS, L-BFGS, BFGS-TRM, CG on different potentials
   - Instanton geometry, hessian and energy calculations

I deactivated the NEB routine until further bugfixes are conducted as its results could not be reproduced by independent tests.


litman90:
I merged the workflow update a few mins ago. Plase, merge master into this branch to resolve the existing conflicts.

ipi_tests/regression_tests/README
```diff
@@ -1,15 +1,33 @@
...
 
    "python test_run.py -h"
 
 To add a new regression test please provide:
 
    - input.xml (and all the files required by it)
 
-   - driver.txt specifing  "<driver> <address> <port> <unix/inet> " (see existing examples)
+   - driver.txt specifying  "<driver> <address> <port> <unix/inet> " (see existing examples)
+	you should run the driver with the command
```
litman90, line 17: what do you mean? 
Shoudn't be something like " the driver will be called using the following syntax ..."

ipi_tests/regression_tests/README
```diff
@@ -1,15 +1,33 @@
...
+ 
+	"i-pi-driver -m <driver> -h <address> -u -o <possible options>" if unix 
+
+   
+   - file_to_check.txt specifying the files serving as reference with their
+     respective filenames (see existing examples)
```
litman90, line 23: Could you point to an example where the "possible options" flags are actually used? Otherwise one has to look for it along the many reg-test

litman90 (commented):
Looks really good.
Please address the few comments I wrote and we merge it today.

ipi_tests/regression_tests/README
eszterpos: Yes, you are totally right, I got a little confused with the wording and was addressing the creation of the references.

ipi_tests/regression_tests/README
eszterpos: yes, done.

litman90 (approved):

PR #133, opened 2020-09-24
Workflow update

eszterpos:
This pull request contains a workflow modification to set up the environment with conda and use different python versions for numerical checks: 3.7, 3.8.5.

The files that have changed are:
  ./.github/workflows/pytest.yml
  ./requirements.txt.

@litman90 : Please, replace the current workflow file on github with pytest.yml (I think the requirements.txt file is fine, the workflow should see the updated version already.


requirements.txt
```diff
@@ -1,4 +1,7 @@
+mkl
+blas=*=mkl
 numpy
+scipy
```
litman90, line 4: Scipy is only used for a few algorithms and it is not present in some clusters,
So we don't want to put it as a requirement.
However, since we want to use it for the regtests we have to install it in as it was done before.

Please take 'scipy' out from the "requirements.txt" and install it in pip

.github/workflows/pytest.yml
```diff
@@ -10,27 +10,38 @@ jobs:
...
+    - name: Test Conda 
+      run: python -c 'import numpy; import sys; print(sys.version); print(numpy.__version__)'
+
+#    - name: Install dependencies with pip
+#      run: |
+#        python -m pip install --upgrade pip
+#        pip install -r requirements.txt
+#        pip install .
+#        pip install scipy
+#        echo "::add-path::${GITHUB_WORKSPACE}/bin"
```
litman90, line 44: Please delete these comented lines

.github/workflows/pytest.yml
```diff
@@ -10,27 +10,38 @@ jobs:
...
 
-    - name: Install dependencies
+    - name: Install dependencies with conda
       run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements.txt
-        pip install .
-        pip install scipy
-        echo "::add-path::${GITHUB_WORKSPACE}/bin"
+         conda install --file requirements.txt
```
litman90, line 31: I know this was done before, but I think the correct way to go is to use conda install .  (in the ipi directory)
And this will install the dependencies automatically

requirements.txt
eszterpos: done so... removed scipy from requirements.txt

.github/workflows/pytest.yml
eszterpos: I install  I-PI as the following:
conda install -f requirements.txt
conda install pip
pip install .

litman90: ok

litman90 (approved):

litman90 (approved):

PR #151, opened 2021-01-09
Tests enhancement. 

ceriottm:
I would like to use this PR to improve the execution of tests, particularly those on examples. 
Work in here does already quite a bit:

- Introduces a driver -m dummy mode that returns random forces, that should be OK to test *any* example
- reduces the time it takes to run an example test by avoiding the initial 3s sleep and by cutting down on the softexit latency, which should be generally a good thing anyway


ceriottm:
Up for discussion: 

- is the .driver_info.txt machinery needed? should we remove it altogether or keep it/transform it?
- how to give better error messages when an example fails.

ipi/engine/motion/dynamics.py
```diff
@@ -607,7 +607,7 @@ class NPTIntegrator(NVTIntegrator):
     def pstep(self, level=0):
         """Velocity Verlet monemtum propagator."""
 
-        if np.allclose(self.forces.vir, np.zeros(len(self.forces.vir))):
+        if np.array_equiv(self.forces.vir, np.zeros(len(self.forces.vir))):
```
litman90, line 610: is this still required if we are using the dummy driver?

ipi/engine/motion/dynamics.py
```diff
@@ -745,7 +745,7 @@ class SCNPTIntegrator(SCIntegrator):
...
```
litman90, line 748: same here

drivers/driver.f90
```diff
@@ -154,9 +155,11 @@ PROGRAM DRIVER
                   vstyle = 24
                ELSEIF (trim(cmdbuffer) == "gas") THEN
                   vstyle = 0  ! ideal gas
+               ELSEIF (trim(cmdbuffer) == "dummy") THEN
+                  vstyle = 99 ! returns non-zero but otherwise meaningless values
                ELSE
                   WRITE(*,*) " Unrecognized potential type ", trim(cmdbuffer)
-                  WRITE(*,*) " Use -m [gas|lj|sg|harm|harm3d|morse|zundel|qtip4pf|pswater|lepsm1|lepsm2|qtip4pf-efield|eckart|ch4hcbe|ljpolymer|MB|doublewell|doublewell_1D] "
+                  WRITE(*,*) " Use -m [duymmy|gas|lj|sg|harm|harm3d|morse|zundel|qtip4pf|pswater|lepsm1|lepsm2|qtip4pf-efield|eckart|ch4hcbe|ljpolymer|MB|doublewell|doublewell_1D] "
```
litman90, line 162: dummy should also be added below (line 699)

ipi_tests/examples/exampletools.py
```diff
@@ -183,7 +183,17 @@ def run(self, cwd, nid):
...
+            if len(clients) > 0:
+                f_connected = False
+                for i in range(50):
+                    if os.path.exists("/tmp/ipi_" + clients[0][1]):
+                        f_connected = True
+                        break
+                    else:
+                        time.sleep(0.1)
+                if not f_connected:
+                    raise RuntimeError("Couldn't find the i-PI UNIX socket")
```
litman90, line 196: The clients are called afterward,
I don't get what are these lines for

litman90:
> Up for discussion:
> 
>     * is the .driver_info.txt machinery needed? should we remove it altogether or keep it/transform it?
> 
>     * how to give better error messages when an example fails.

1) Maybe transforming it to .test_info.txt where we can provide more data, like the number of steps, would help to cover the corner cases which are not covered by the default setup, like imf which requires 7 steps.

2) are you referring here in the CI, or locally when calling the test_examples.py script? 
Could you provide/describe one case for which you think we can do better?

ipi/engine/motion/dynamics.py
ceriottm: This actually covers a safety net that @mahrossi put in place for ppl doing NPT without activating the calculation of the stress. I was bugged that this broke the ideal gas driver, so I had gas return 1e-100 and made these check for *exactly* zero

ipi_tests/examples/exampletools.py
ceriottm: This substitutes the `sleep(3)`.  It waits until i-PI has created the sockets before letting the clients start.

ceriottm:
> > Up for discussion:
> > ```
> > * is the .driver_info.txt machinery needed? should we remove it altogether or keep it/transform it?
> > 
> > * how to give better error messages when an example fails.
> > ```
> 
> 1. Maybe transforming it to .test_info.txt where we can provide more data, like the number of steps, would help to cover the corner cases which are not covered by the default setup, like imf which requires 7 steps.

I agree, but while we are at it I'd make the file visible. Since these will be used *very* sparingly (ideally the heuristics should be smart enough not to need instructions) I think it's best to make them visible - perhaps with a comment explaining what the file is for. 

> 2. are you referring here in the CI, or locally when calling the test_examples.py script?
>    Could you provide/describe one case for which you think we can do better?

When calling test_examples.py it only tells you that the example failed. It doesn't give errors, nor there are error logs in the run folder. Incidentally, when an example runs successfully it'd be good to clear the tmp folder, running the full thing generates a gigabyte of crap, and it adds up if you run the examples multiple times.

eszterpos:
> > > Up for discussion:
> > > ```
> > > * is the .driver_info.txt machinery needed? should we remove it altogether or keep it/transform it?
> > > 
> > > * how to give better error messages when an example fails.
> > > ```
> > 
> > 
> > 
> > 1. Maybe transforming it to .test_info.txt where we can provide more data, like the number of steps, would help to cover the corner cases which are not covered by the default setup, like imf which requires 7 steps.
> 
> I agree, but while we are at it I'd make the file visible. Since these will be used _very_ sparingly (ideally the heuristics should be smart enough not to need instructions) I think it's best to make them visible - perhaps with a comment explaining what the file is for.
> 
> > 1. are you referring here in the CI, or locally when calling the test_examples.py script?
> >    Could you provide/describe one case for which you think we can do better?
> 
> When calling test_examples.py it only tells you that the example failed. It doesn't give errors, nor there are error logs in the run folder. Incidentally, when an example runs successfully it'd be good to clear the tmp folder, running the full thing generates a gigabyte of crap, and it adds up if you run the examples multiple times.

mahrossi:
this is really still WIP or should we move on here too?

ceriottm:
Depends on how much we want to do here. Personally I'd like to remove all the .driver_info files, and improve the error messages you get when running test_examples.py directly. 
Also, if - following Yair's suggestion - we use an explicit test-settings.txt file to deal with some of the most annoying special cases, I would do the renaming on the code side here, and then the actual fixing of the tests in the zero-errors branch.

litman90:
I am in favor of extending  .driver_info to a more general test-settings.txt file, and of course, improving the error messages always helps.
On this front, in the PR #155 pytest is called with some flags to get a nice short summary report at the
end about the errors and failures.

ceriottm:
Since we agree on how to proceed with driver_info, do you want to do that? You know the test infrastructure much better than I do ^_^
Re errors, I guess what I found confusing is that the README files in ipi_tests encourages to run the test_XXXX.py files manually. If the error reports are more telling when run through pytest (they indeed seem to be) it would be better to indicate how to run individual examples using pytest. So perhaps this is just a matter of updating the READMEs?

litman90:
Probably an update on the README file together with some cleaning up will be best. I can do it next week.
Re the more general test-settings.txt, besides specifying the driver and the number of steps, what else should be considered?

ceriottm:
> Probably an update on the README file together with some cleaning up will be best. I can do it next week.
> Re the more general test-settings.txt, besides specifying the driver and the number of steps, can you imagine something else that should be considered?

I would leave all the options that are there already, but make it clear in the README that this should be unnecessary for the vast majority of cases. And add functionality for comments in the settings files, which I don't know if it's there already.

litman90:
The goal of this PR benefits partially from the PRs 154,155 and 156. There are still a few things to improve. I was planning to take care of the missing features today and Monday, or is @mahrossi working here?

ceriottm:
can we merge this?

litman90 (approved):

ceriottm:
ah @litman90  sorry I had missed your message from earlier on. thanks for merging this in

PR #153, opened 2021-01-11
Black files that were not black 20.8b compliant

litman90:
Some files were still no black 20.8b compliant.


litman90:
This PR changes only a few files that were not black 20.8b compliant. 99.99% Harmless.

PR #154, opened 2021-01-11
Split tests

litman90:
This PR split the unitary and regtests from the "examples" tests.

The idea is that we trigger  unitary and regtests in each push and
all the test when doing a PR to master.


litman90:
I also simplified the regression tests for the instanton functionality because they were taking an unnecessarily long time.

bin/i-pi-tests
```diff
@@ -0,0 +1 @@
+../tools/py/tests.py
```
mahrossi, line 1: is this the most general way to link these files? why not making it a link to tests.py instead?

litman90: I did a symbolic link  (ln -s ) like it is done for all the files in the bin folder except for "i-pi"
What are you suggesting?

mahrossi: ah is this how a symbolic link appears in github? interesting, I did not know! sorry, I only looked at the review interface and not the command line. then you did what I suggested.

mahrossi (approved):

eszterpos:
This is great, and will make the pushes considerably a smoother experience for the developers.

PR #155, opened 2021-01-12
Clean workflow

litman90:
Now the linting is done by 'black . --check' which is consistent with using black as a format tool.
I also separated the lint workflow from the other tests so when the linting fails, the other tests are not aborted


ceriottm:
Looks OK but shouldn't we now remove all references to flake8 if we use black for checking?

ceriottm:
Incidentally I think it's urgent to merge this branch, because the linting tests seem not to run in all the branches that are synced to master.

eszterpos:
I have added flake8 back while limiting its use to act only as a syntax linter, and fixed some warnings in the code, it was raising before (mainly about not used variables). I agree that it would be great to merge this soon, so we could update the workflows on github and the other PR-s can pass.

litman90:
It looks good to me.  @mahrossi or @ceriottm could you check and if you find it also ok merge it?

PR #158, opened 2021-01-19
Rename workflows

litman90:
This PR renames the workflows with a more intuitive name, without changing what is being tested.


eszterpos:
Great and more logical naming of workflows. Now the different checks for examples, linting and regtests are separated which eases the possible debugging issues.

PR #159, opened 2021-01-20
Python 3.9 compatibility by replacing deprecated Thread.isAlive() with Thread.is_alive()

eljost:
Dear ladies and gentlemen,
currently I-PI is not compatible with Python 3.9+, as in 3.9 Thread.isAlive() was finally [deprecated](https://bugs.python.org/issue37804) in favor of Thread.is_alive(). This (very) small PR updates the four remaining occurrences of isAlive() to is_alive().

All the best,
Johannes


mahrossi (approved):
Looks good to me and it is also back-compatible with python3.7 -- so, thanks a lot!

PR #161, opened 2021-01-26
Memory leak fix and debug infrastructure

ceriottm:
Spotted a bad memory leak while polishing the npzt branch. 
I also added some code to detect future memory leaks, it prints statistics about memory allocations when running with verbosity=debug.


mahrossi (commented):
Can't see any problems and it was obviously debugged by MC so... -- Side note: A student here saw traces of what I think must be this memory leak when running NPT with many beads for a long time. I had asked him to try to narrow it down and see if it was really a leak but never heard back :-D

mahrossi (approved):

PR #162, opened 2021-01-26
Python driver

ceriottm:
I created a minimal driver.py that can be used as a template by others wanting to create a simple interface to i-PI.
Perhaps we can also consider it as an alternative way to close #107 
We are currently using this with librascal so if we process the PR soon we make sure there's no divergence.


drivers/driver.py
```diff
@@ -0,0 +1,169 @@
...
+                # shapes up the position array
+                pos.resize((nat, 3))
+                force.resize((nat, 3))
+            else:
+                if len(pos) != nat:
+                    raise RuntimeError("Atom number changed during i-PI run")
+            pos = recv_data(sock, pos)
+
+            ##### THIS IS THE TIME TO DO SOMETHING WITH THE POSITIONS!
+            pot, force, vir, extras = driver(cell, pos)
```
litman90, line 123: shouldn't this be dummy_driver instead of driver?

litman90 (commented):
I suggest coding a simple driver and compare with the Fortran driver to catch most of the trivial mistakes before merging.

drivers/driver.py
ceriottm: ah crap this is confusing  I agree

ceriottm:
Coded a harmonic driver and checked it is identical to the fortran driver. IDK if either @litman90  or @eszterpos want to add some kind of regtest - I think it'll take you a fraction of the time it'd take to me. Otherwise I'd merge so we have one less PR open.

litman90:
I will implementing few new python drivers next week. I can create the corresponding regtest, either in this branch or in a new PR.

ceriottm:
> I will be implementing a few new python drivers next week. I can create the corresponding regtest, either in this branch or in a new PR.

Might be good to make a separate PR - also to keep driver.py clean for people who want to just port it to their own project (we have our own baking into librascal) it might be good to keep driver functions in a separate file. I will create it already in this PR.

litman90:
I upgraded the dummy driver to a class and added the possibility to pass parameters to the driver. 
The harmonic driver now accepts a user-defined force constantly.
I tested against the FORTRAN driver, and everything looks fine
If @ceriottm is fine with these changes, and all the test pass, it is ready to merge.

PR #163, opened 2021-01-26
Extras tweak

eszterpos:
This PR is there to tweak and update some inaccuracies around the extras passed as a JSON formatted string and its handling later in the code.

There is an update about the formatting of the friction tensor being printed and corresponding updates in the regtest checking friction tensor printing.


ceriottm (dismissed):
Have gone through the whole PR offline, and agreed several changes with @eszterpos. Seems good to go as far as I'm concerned, we also cleared a lot of ancient poor choices of docstrings and comments

examples/pes/pswater/contract-extras/input.xml
```diff
@@ -0,0 +1,47 @@
+<simulation verbosity='medium'>
```
mahrossi, line 1: I don't think this needs to be addressed in this PR, but just to point out that this is another example of a "examples"  structure where finding the "contract-extras" functionality will be hard (who will imagine to look inside pswater). I will try to propose a reorganization.

examples/pes/pswater/nvt/input.xml
```diff
@@ -5,7 +5,9 @@
     <trajectory filename='vel' format='xyz' stride='5'> velocities </trajectory>
     <trajectory filename='for' format='xyz' stride='2'> forces </trajectory>
     <trajectory filename='kin' format='xyz' stride='2'> kinetic_cv </trajectory>
-    <trajectory filename='dip' stride='2'> extras </trajectory>
+    <trajectory filename='dip' stride='2' extra_type='dipole'> extras </trajectory>
+    <trajectory filename='raw' stride='2' extra_type='raw'> extras </trajectory>    
+    <trajectory filename='dip0' stride='2' extra_type='dipole'> extras_component(0) </trajectory>
```
mahrossi, line 10: Also here, can we be a bit more verbose about what this does, at least in the examples input file? And also here the example is even harder to find :-D

ipi/engine/forcefields.py
```diff
@@ -631,7 +631,7 @@ def evaluate(self, r):
         v = bias[0]
         vir *= -1
 
-        r["result"] = [v, f, vir, ""]
+        r["result"] = [v, f, vir, {"raw": ""}]
```
mahrossi, line 634: The "raw" field is for the extras?

ipi/engine/forces.py
```diff
@@ -301,6 +307,8 @@ def __init__(
               will be weighted by this factor.
            name: The name of the forcefield.
            mts_weights: Weight of forcefield at each mts level.
+           force_extras: A list of properties that should be treated as physical quantities,
+              converted to numpy arrays and treated with ring polymer contraction
```
mahrossi, line 311: Here a description how to treat it with RP-contract would be good. I can amend this comment in a moment.

ipi/engine/forces.py
```diff
@@ -417,11 +429,30 @@ def extra_gather(self):
...
+        for b in self._forces:
+            for e in b.extra:
+                if not e in fc_extra:
+                    raise KeyError(
+                        "Extras mismatch between beads in the same force component, key: "
+                        + e
+                    )
+                fc_extra[e].append(b.extra[e])
+
+        # force_extras should be numerical, thus can be converted to numpy arrays
```
mahrossi, line 451: "should be"? normally they are and that makes the most sense, but I don't see where the error is raised in a clean way. should be verbose.

ipi/engine/forces.py
```diff
@@ -1331,16 +1377,54 @@ def pot_combine(self):
...
+            # combines the extras from the different force components
+            for e, v in self.mforces[k].extras.items():
+                if e in self.mforces[k].force_extras:
+                    # extras that are tagged as force_extras are treated exactly as if they were an energy/force/stress
+                    v = (
+                        self.mforces[k].weight
+                        * self.mforces[k].mts_weights.sum()
+                        * self.mrpc[k].b2tob1(v)
+                    )
+                    if e in re:
```
mahrossi, line 1395: Uhh.... I thought we had agreed that silently summing everything was not a good idea. Why is this now being done regardless?

ipi/engine/outputs.py
```diff
@@ -273,7 +272,7 @@ def __init__(
         format="xyz",
         cell_units="atomic_unit",
         ibead=-1,
-        xtratype="info",
+        extra_type="raw",
```
mahrossi, line 275: I see what raw is, thanks. The comment above is obsolete.

ipi/engine/outputs.py
```diff
@@ -290,7 +289,7 @@ def __init__(
            cell_units: A string specifying the units that the cell parameters are
               given in.
            ibead: If positive, prints out only the selected bead. If negative, prints out one file per bead.
-           xtratype: Specifies the type of extras string that is printed in the file
+           extra_type: Specifies the type of extras string that is printed in the file
```
mahrossi, line 292: Should we add that type is used-defined?

mahrossi (commented):
Beyond my minor comments, I will still run a few examples also of old functionalities with extras to be sure that all is fine. As the "biggest" comment I would say that this is now a quite complex (and good, and necessary) functionality that has extremely little documentation at the moment. This could be better addressed.

examples/pes/pswater/contract-extras/input.xml
ceriottm: I had an idea about this yesterday. We need a script in examples/ that looks for inputs that use a given keyword.
I always go there and `grep` the recursive shit out of the folder. Automating and documenting this will be very useful IMHO.

ipi/engine/forces.py
ceriottm: Counting on np.asarray() to raise an error.

ipi/engine/forces.py
ceriottm: So, the rationale here is that quantities that are "promoted" to force_extras status are treated as if they were forces. This means they can be contracted, combined, summed and weighted. I can see more cases in which this is useful than cases in which it is a problem: if you don't want your extras to be summed, you can (a) not include them in force_extras or (b) give them a different name in each forcefield.

ceriottm: So essentially the default is "don't touch the extras". If you activate the force_extras machinery, then they get summed and resampled.

ipi/engine/outputs.py
ceriottm: why not. and that it defaults to printing the full "raw" string

ipi/engine/forces.py
mahrossi: ok, let's raise an actual error to help the users...

mahrossi (dismissed):
Ok I am happy with it and we ironed-out all minor points.

mahrossi (approved):
Still approved

PR #167, opened 2021-02-12
Setup documentation using Sphinx

Luthaf:
I also imported the existing manual (except the input reference) using [pandoc](https://pandoc.org/) to convert latex to rst, this was surprisingly easy.

The remaining items to complete the documentation are
1) Convert create_man to output RestructuredText instead of LaTeX
2) Write up the FAQ

This is a draft PR since I still need to test the deployment on github pages, and add a pdf builder as well.


Luthaf:
This seems to be working, you can see the output here: https://i-pi.github.io/i-pi/. The PDF version is available at https://i-pi.github.io/i-pi/ipi-manual.pdf

doc/src/conf.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = "i-PI"
+copyright = "2021, The i-PI developers"
+author = "The i-PI developers"
+
+# The full version, including alpha/beta/rc tags
+release = "TODO VERSION"
```
Luthaf, line 25: sphinx expects a version number, what should I put here?

PR #170, opened 2021-02-17
First entries in doc/src/faq.rst

fidanyan:
Some FAQs based on the forum questions


mahrossi:
could you add some points about usage of i-PI in HPC infrastructures? where to find information and basic stuffs?

another point is about units in i-PI -- lots of people get errors because of that

fidanyan:
I see that none of the examples has a sample of slurm script to run it on a cluster. I think I'll add it to `lammps/benz-graph-slc` and then will refer to it in the FAQ and in the manual.

fidanyan:
Another small thing: I found outdated reference in the latex manual: page 6 `examples/pes-regtest/io-units/`.
I changed it to `examples/lammps/*/init.xyz` and added `examples/qespresso/h2o/init.pdb`, because PDB is not covered otherwise.

examples/slurm/README
```diff
@@ -0,0 +1 @@
+This folder gives examples of setup to run i-PI on HPC systems.
```
mahrossi, line 1: I am unsure about the usefulness of this full example. The slurm submission script has lots of useful bits, no doubt -- but is the whole tree useful?

doc/latex/manual.tex
```diff
@@ -48,7 +48,7 @@
 \hbox{}
 \vspace*{\fill}
 \begin{center}
-\
```
mahrossi, line 51: Lots of white spaces were removed and that is why the manual seems so changed.

examples/slurm/README
fidanyan: The idea was to have a truly complicated setup done from the beginning to the end, so that person can look ar everything together. It could be smaller in case of driver clients, but I didn't have anything with the driver that I could use for SL-RPC.

mahrossi: ok! could we at least do a symbolic link for the CH.airebo file and for the data.*** files instead of recopying everything?

fidanyan: Yes, I can in principle link all such files to `lammps/benz-graph-slc/`. I was thinking about making this only an "addon" to benz-graph-slc example, but decided that it will be unclear for newcomers compared to a standalone example. Linking also complicates things, but it's a reasonable compromise.

mahrossi: yep just do it

fidanyan: Done.

mahrossi (approved):

PR #175, opened 2021-02-21
Force match mode

Taylor-96:
An attempt at implementing a matching mode where drivers are consistently assigned to a given replica.

### DIFF
--- a/ipi/interfaces/sockets.py
+++ b/ipi/interfaces/sockets.py
@@ -764,6 +764,8 @@ class InterfaceSocket(object):
             match_seq = ["match", "none", "free", "any"]
         elif self.match_mode == "any":
             match_seq = ["any"]
+        elif self.match_mode == "force_match":
+            match_seq = ["match", "none"]
 
         # first: dispatches jobs to free clients (if any!)
         # tries first to match previous replica<>driver association, then to get new clients, and only finally send the a new replica to old drivers
@@ -778,6 +780,11 @@ class InterfaceSocket(object):
 
                     if len(self.prlist) == 0:
                         break
+            # if using forced match mode, check that there is a at least one client-replica match in the lists freec and prlist.
+            #If not, we break out of the while loop
+            if self.match_mode == "force_match":
+                  break
+          
             if len(freec) > 0:
                 self.prlist = [r for r in self.requests if r["status"] == "Queued"]
         tdispatch += time.time()
@@ -837,6 +844,11 @@ class InterfaceSocket(object):
                 continue
             elif match_ids == "none" and fc.lastreq is not None:
                 continue
+            elif self.match_mode == "force_match" and match_ids == "none" and (r["id"] in [c.lastreq for c in self.clients]):
+                # if using forced match mode and the user connects more clients than there are replicas, do not allow this client to
+                # be matched with a pending request. 
+                continue
+
             elif match_ids == "free" and fc.locked:
                 continue
 
--- a/ipi/interfaces/sockets.py
+++ b/ipi/interfaces/sockets.py
@@ -781,10 +781,10 @@ class InterfaceSocket(object):
                     if len(self.prlist) == 0:
                         break
             # if using forced match mode, check that there is a at least one client-replica match in the lists freec and prlist.
-            #If not, we break out of the while loop
+            # If not, we break out of the while loop
             if self.match_mode == "force_match":
-                  break
-          
+                break
+
             if len(freec) > 0:
                 self.prlist = [r for r in self.requests if r["status"] == "Queued"]
         tdispatch += time.time()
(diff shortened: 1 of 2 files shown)


ipi/interfaces/sockets.py
```diff
@@ -837,6 +844,15 @@ def dispatch_free_client(self, fc, match_ids="any", send_threads=[]):
                 continue
             elif match_ids == "none" and fc.lastreq is not None:
                 continue
+            elif (
+                self.match_mode == "force_match"
+                and match_ids == "none"
+                and (r["id"] in [c.lastreq for c in self.clients])
+            ):
+                # if using forced match mode and the user connects more clients than there are replicas, do not allow this client to
```
mahrossi, line 852: This could result in not-ideal situations for the user, in which they unknowingly have a client just idling. How is this situation handled?

mahrossi:
Hi Taylor, thanks for starting this! It would be very helpful if you could build an example using either the python driver or the fortran driver that live inside the `drivers` folder in i-PI. You can create an example in the `examples/pes` folder that showcases this mode of operation and will also help us test all possible scenarios quickly. If you need help with that Slack us at any time.

ipi/interfaces/sockets.py
Taylor-96: Hi, thanks for the feedback. I'll have a go building a test using one of the driver codes - will let you know in Slack if I run into any difficulties. Regarding the potential idling of excess drivers, I was actually wanting to ask yourselves how you'd like this to be handled. I wondered about forcefully terminating that superfluous driver by passing it an exit signal. However, I was unsure if something more gentle like a simple warning message would be preferable so that the user can handle shutdown of the driver as they wish?

ceriottm:
A thought. Rather than calling this mode "force_match", which is confusing as (1) it's also the name of a coarse-graining approach and (2) is too similar to the "match" approach, how about we call this "lock"?

Taylor-96:
Hey, yeah I was thinking force_match sounds a little too like asking forces to be matched or something (wasn't aware of the coarse-graining method though). "Lock" sounds good! I'll make the switch.

Taylor-96:
Hi all, after speaking with Michele last week, I'm just resurrecting this PR. I just added an example (in examples/pes/lock_mode) which is essentially the 4 replica harmonic example but now instantiating 4 drivers in the Makefile rather than 1 (with lock_mode the simulation would hang otherwise).
I also tried to include a regression test for lock_mode. I was a bit uncertain how to go about this based on the way the other tests were structured. I based the test on the 2 bead NVE test (in ipi_tests/regression_tests/tests/NVE/NVE_2/) but again I need to instantiate 2 separate drivers in the case of lock_mode. Looking at test_tools.py, it seemed like I should specify two separate drivers in test_settings.dat. Unfortunately for some reason I cannot run the tests on my local machine either using i-pi-tests or run_tests.py (they just hang for me even with a fresh i-pi installation) and so I tried to run the tests on a remote machine. They appeared to work as intended but apologies if this isn't the case. 

Regarding the point you raised about idling of superfluous drivers Mariana, if you think it would be best to have these be forcefully shutdown I can look back into how this is done (my memory is a bit hazy regarding how this is done since it has been a while since I looked at the code). 
An issue I could envision arising for people who want to use this mode of running when employing a driver code that requires consistency of the replica<->driver matchup may be in restarting their simulations. I will likely encounter this problem soon and was going to address it on the driver side by saving electronic densities in driver subdirectories together with the associated replica ID so that the match up between electronic state/nuclear geometry can be carried out at restart. Not sure if it is something you feel could be useful to have functionality for on the server side, however. 

Hope this all looks okay. Cheers!

mahrossi:
Hi Taylor, great, thanks for resurrecting it - quite amazing job. If the tests pass the PR, all should be good. If you think you could add a "Warning" that users should check whether they have idling clients when using this mode, that could be a first-step solution. Then when you encounter problems or find solutions, please report back and we can tune the functionality. We will not leave any PR idling for too long now - really sorry for this one.

mahrossi:
Hey @Taylor-96 the tests are not passing for your new lock-mode example and regtest. I think it is a problem with how you defined / did these tests. The rest looks good. Did they pass in your machine?

Taylor-96:
So I think I've managed to get the regtest working now too if I modify modify_xml_4_dummy_test further to account for the driver flags: 
```python 
  #if there are remaining drivers, assign them to the last socket
        if ((s==len(ff_sockets) -1) and (len(driver_info["driver_model"]) > len(ff_sockets))):
            for remaining_client_idx in range(s+1,len(driver_info["driver_model"])):
                model = driver_info["driver_model"][remaining_client_idx]
                clients.append([model, "unix", address, port])

                for key in driver_info["flag"][remaining_client_idx].keys():
                    if "-" in key:
                        for k, v in driver_info["flag"][remaining_client_idx].items():
                            clients[remaining_client_idx].append(k)
                            clients[remaining_client_idx].extend(v)
````
If these modifcations seem alright I can push again. Not sure how to have property testing of "lock" mode without these changes to the test scripts.

mahrossi:
Thanks @Taylor-96 I'll have a check whether these modifications don't break anything else. Please push all modifications if you haven't yet (although it seems like you did in the last commit :-D) we test before merging.

mahrossi:
So yes, I think your modification to the driver script will not break anything and I understand it because I know the problem you were solving - also I think it is good to test the connection of more than one client to the same socket. I will approve this and merge.

mahrossi (approved):
Looks good and adds possibility of testing connection of multiple clients. Maybe code can be made more transparent for that part of the script at a later iteration.

Taylor-96:
Hi, that's great! Glad the modifications can come in handy in the future! Thanks again for helping me with this!

PR #176, opened 2021-02-26
Fixing init.pdb isssue

### LINKED ISSUE #173: Examples inputs: PDB + repo bloat
We noticed many examples (mostly water) use an input.pdb initial structure that is in non-standard (atomic units) format. Would be good to change all to use a xyz, or angstrom-units pdb. 
Also, since many examples share the same inputs, it might be nice to re-use the same input, perhaps having a common "init/" folder at the top level of examples. Possibly this might break current example runner, though.

LyKex:
fixes #173


mahrossi:
I think this PR should only be merged when all the pdb issues are solved and we find a way to get rid of them from the history. At the moment this does not yet fixes #173.

ceriottm:
Of course. It will fix once it's finished ^_^

LyKex:
Hi, just wondering what is the naming rule for the init files? 
What does 216 stands for in `liquid_water_216.xyz `?

ceriottm:
there are no set naming rules -- I just renamed that to be descriptive,
it's a box of 216 water molecules in the liquid phase.

ceriottm:
I actually think it would be nice to leave the name of the symlink be equal to the one in the init_files folders. It's nicer to read that we initialize from `water_300K_216.xyz` rather than `init.xyz`. 
All agree? Also, I wouldn't make statements implying these are equilibrated at 300K or 200K. Just call them "water_216.xyz" and "ice_96.xyz"

LyKex:
One of the newly added regtest failed to run. `NVE/NVE_1/harmonic_python`
 Is it related to the #181 ?

ceriottm:
merged the librascal branch, regtests pass. what else is there to be done? time for review methinks

eszterpos:
The PR changes the setup of examples and regtest init files to move from the pdb infrastructure to the xyz infrastructure as an important first step in de-bloating the repo. All the tests are passing and as I don't see any problems with the changes implemented so I approve it.

eszterpos (approved):
approving

PR #177, opened 2021-02-26
Organization of drivers folder that got a bit crowded with the last merge

mahrossi:
Here I created a "fortran" and "python" folder inside the original "drivers" folder. I then moved all relevant files inside each. I also renamed the "drivers/drivers" folder (a bit recursive) to drivers/python/pydrivers. I tweaked makefile and driver.py for consistency. Should be a straightforward merge.


mahrossi:
Ah well, I was wrong and it got too late. Some path/link in the examples-tests // tools still needs to be updated. Try to tackle it with more patience some other time.

mahrossi:
Thanks @ceriottm for the further tweaks - there was only a link missing to make it all work! Now all tests pass!

mahrossi:
A comment to @litman90 and @eszterpos -- the CI was taking ~1.5 min until last PR #174 -- now it is taking 10 min, 6 of which go into the testing of examples. I think something was added there that is doing way more than a few tests. This is not urgent but should be addressed.

mahrossi:
A comment regarding #174 and documenting the creation of examples and regtests (@litman90 @eszterpos @venkatkapil24 ) -- forgetting to add "nsteps" to "test_settings.dat" can increase the time of the CI a lot. Need to document that carefully. I fixed two instances that were created in #174 but not sure they are all.

ceriottm:
> A comment regarding #174 and documenting the creation of examples and regtests (@litman90 @eszterpos @venkatkapil24 ) -- forgetting to add "nsteps" to "test_settings.dat" can increase the time of the CI a lot. Need to document that carefully. I fixed two instances that were created in #174 but not sure they are all.

looking at the code, the default nstep should be 1. weird. still, this does what it says on the envelope, and doesn't increase the time beyond what's happening on other branches synced to master, so I think this can be merged.

PR #178, opened 2021-03-02
Changes to support generation of a package for PyPI

### LINKED ISSUE #16: Anaconda cloud package
Would there be an interest to have i-PI available on the Anaconda cloud? I'd be happy to setup a CI on Travis that would build the packages for macOS and Linux and deploy to the cloud.

jhrmnn:
Would fix #16

### DIFF
--- a/ipi_tests/unit_tests/utils/io/backends/test__init__.py
+++ b/ipi_tests/unit_tests/utils/io/backends/test__init__.py
@@ -7,7 +7,7 @@ from pytest_mock import mocker
 import numpy as np
 import numpy.testing as npt
 
-import ipi_tests.unit_tests.common.xyz_generator as xyz_gen
+from ....common import xyz_generator as xyz_gen
 
 import ipi.utils.io as io
 import ipi.utils.mathtools as mt
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,34 @@
+[metadata]
+name = i-PI
+version = 2.3
+description = A Python interface for ab initio path integral molecular dynamics simulations
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+author = The i-PI developers
+author_email = ipi.managers@gmail.com
+url = http://ipi-code.org
+project_urls =
+    Documentation = https://ipi-code.org/i-pi
+    Repository = https://github.com/i-pi/i-pi
+license_files =
+    licenses/license_GPL.txt
+    licenses/license_MIT.txt
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Environment :: Console
+    Intended Audience :: Science/Research
+    Operating System :: MacOS :: MacOS X
+    Operating System :: POSIX :: Linux
+    License :: OSI Approved :: GNU General Public License v3 (GPLv3)
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Topic :: Scientific/Engineering :: Chemistry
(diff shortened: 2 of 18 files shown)


ceriottm:
works as advertised. weird regtest failure which might be more of our problem that this PR's. @litman90 any thoughts?

setup.cfg
```diff
@@ -0,0 +1,38 @@
+[metadata]
+name = i-PI
+version = 2.3
+description = A Python interface for ab initio path integral molecular dynamics simulations
+long_description = file: README.rst
+long_description_content_type = text/x-rst
+author = Michele Ceriotti
```
mahrossi, line 7: Maybe the authors could be ammended :-) i-PI development team and the e-mail of i-PI coordinators?

mahrossi:
This failing in the regtest has been observed many times before, I believe mainly by @eszterpos -- if I recall correctly it has to do with the BLACS/MKL that is being linked to numpy. But I cannot see what changed in this commit to make that come back. Need to look deeper.

setup.cfg
jhrmnn: Sure, would that be ipi.managers@gmail.com?

mahrossi: yes (note to self we need to add more people to this alias)

jhrmnn:
Ah, I marked this as a draft PR, so this wasn't supposed to be merged yet (need to update the email, and add dependencies), but nevermind, I'll make another one.

jhrmnn:
The second commit is a bit more invasive, so another review is welcome.

I've included `drivers/py` and `tools/py` as `ipi._driver` and `ipi._tools` subpackages, respectively. This means one can run these as `python -m ipi._driver.driver --help` or `python -m ipi._tools.committee_reweight --help`. This is achieved by standard arguments to `setup()`.

I went a bit further and wanted to imitate the symlinked scripts in `bin/`, so these are first listed in `setup.py` and then delegated to a new module `ipi._launcher`, which handles them. There is a bit of import magic involved, but then again i-PI has no ordinary directory structure...

ceriottm:
@mahrossi I think it would be appropriate to write "The i-PI developers" as
author.

jhrmnn:
> I think it would be appropriate to write "The i-PI developers" as author.

Done.

jhrmnn:
Ping, can we merge this and go forward with the pypi package?

ipi_tests/unit_tests/utils/io/backends/test__init__.py
```diff
@@ -7,7 +7,7 @@
 import numpy as np
 import numpy.testing as npt
 
-import ipi_tests.unit_tests.common.xyz_generator as xyz_gen
+from ....common import xyz_generator as xyz_gen
```
mahrossi, line 10: If we restructure any of the folder tree here, all of this will have to be addressed, am I right? just a question, but still...

mahrossi (approved):
Ok by me. A bit nervous about all the relative paths for importing stuff in the testing/regtest scripts. But I guess there is no other way, at the moment, to solve that for the packaging?

ipi_tests/unit_tests/utils/io/backends/test__init__.py
jhrmnn: That's true. There isn't that many of them, though, and I assume the restructuring wouldn't be done too often, so hopefully it shouldn't be an issue?

jhrmnn:
Yeah, the problem is that when i-PI gets installed, the tests have to be installed under the `ipi` namespace. The installation puts them under `ipi.tests`, and for that to work, the imports need to be relative, because `ipi_tests` doesn't exist in the installed package.

On the other hand, the advantage of relative imports is that as long as you move the common prefix, you don't have to change anything (say, `ipi_tests` to `tests`).

PR #179, opened 2021-03-03
Solving issues with references in the CI

mahrossi:
I slightly relaxed the tolerances for comparison of references in the CI and now the tests pass again, but I am not really happy with this solution. It is practical but may lead to unwanted effects in the future. These criteria have passed in the local branch.


ceriottm:
not fixed yet....

mahrossi:
> not fixed yet....

This is very weird. It passed _in github_ but in my group's repo! One would have to analyze the outputs

PR #187, opened 2021-04-22
Property "bead_potentials" to output all bead pots at once

fidanyan:
New property bead_potentials is introduced, It will output potentials of all beads.
There is a check in engine/outputs.py:PropertyOutput.print_header() for negative "size" of property, or if type(size) is str. If someone implements a new property with dynamical size and forgets to specify this size in Properties.bind(), the code will complain.


fidanyan:
Hello people, I noticed that this feature was never merged to master. Could we try again? :)

mahrossi (approved):
Looks simple and good to me.

PR #189, opened 2021-05-05
update doctstings to new black standards

litman90:
Black stable has been update inside github actions. This PR handles the new standards for docstrings.


eszterpos (approved):
Simple PR just changing the format of the docstrings as the black standards have changed.

PR #195, opened 2021-08-10
DRAFT: Add ase support

stur86:
I'm marking this as still a draft since it still needs some testing and I wanted to ask help on how to do it.

I've implemented reading and writing of structures using the [Atomic Simulation Environment](https://wiki.fysik.dtu.dk/ase/), which in theory means supporting dozens of new file formats at the cost of one single library to install. The dependency is optional (it should print an error message if one tries to use it when lacking, but otherwise doesn't affect the rest of the code). The way to use it is simply to set `mode='ase'` in a `<file>` XML tag. I think this should be really straightforward, but I'm not sure how and where to implement tests (I'm a bit confused by how the `from ..common.folder import local` works in `test_io.py` since it only seems to contain one XYZ file but is used also for PDB ones), so I was hoping for some help. I'm also not sure about how to test the writing of files, though I've implemented a `print_ase` method too.


stur86:
Hi @ceriottm , are you able to take a look at this at some point? Thanks!

ceriottm:
sorry complicated period. trying to get someone else to have a look. don't lose hope ^_^

stur86:
Any progress on this?

ceriottm:
Ciao, given it seems nobody is looking at this, I'm checking it out. for starters, it'd be enough to add an example folder, e.g
[ase-example.zip](https://github.com/i-pi/i-pi/files/7516135/ase-example.zip)

I can already see an issue: if you open the file from the filename, ASE can guess the filetype from the extension. given we pass the file object, it can't and in this example it cannot even read an xyz.

stur86:
> Ciao, given it seems nobody is looking at this, I'm checking it out. for starters, it'd be enough to add an example folder, e.g [ase-example.zip](https://github.com/i-pi/i-pi/files/7516135/ase-example.zip)
> 
> I can already see an issue: if you open the file from the filename, ASE can guess the filetype from the extension. given we pass the file object, it can't and in this example it cannot even read an xyz.

I fixed this issue in the latest commit - now the reader checks the filename and extracts the extension from there. I don't know if this was necessary as ASE is pretty good at automatic detection and in fact identified the structure just fine even before. The reason for the example's failure is different, and is in the use of `units` and `cell_units`. I don't think these should be allowed at all for this tag. ASE internally uses Angstroms, always; if the file format comes with some other units (and I don't think it actually would work with a XYZ using atomic units, since I believe ASE assumes as part of the format that the units are Angstroms, but each format has its own definition and so ASE can perform conversions in some cases), then ASE should be left to take care of it, and will still return positions and cell in Angstroms. So we should fix the units and remove support for those options in this tag.

Right now, I get a failure with this message:

    TypeError: atomic_units is an undefined unit for kind length.

which is puzzling, and I'm not sure why it happens. But regardless, it's unnecessary and potentially harmful to even allow this for the ASE reader.

stur86:
After checking in depth the ASE code I actually decided to roll back the change. ASE already has a lot of clever detection mechanisms for the file format which works also if you pass a reference, including a more sophisticated version of what I'd added, so I'm only making things worse if I add that. The issue here has definitely more to do with the units.

ceriottm:
OK. Units for files have a tormented history. The original idea was to have a very clear, no-exceptions "units are atomic units unless otherwise specified". Even `<temperature> 300 </temperature>` will be interpreted as 300 Hartree if you don't specify `units='kelvin'`. But with I/O there is the problem of interfacing with standards, and so we made PDB default to angstrom. 
Ideally, I'd like to have a similar arrangement here: defaults to Angstrom, but if one wants to use the ase reader to parse an xyz that is in nanometers, why not? 

So I suggest the following: pls make an example, ideally based on the lammps ice template so the example runs, but using a non-orthogonal unit cell: another issue we'll need to tackle is that i-PI expects structures that have a unit cell with u1//x, u2//xy, u3  generic, so an ASE parser will have to rotate. 

Then I can help you dealing with units conversion if the user does specify them - and perhaps print a warning, I think we do something similar for PDB.

stur86:
> So I suggest the following: pls make an example, ideally based on the lammps ice template so the example runs, but using a non-orthogonal unit cell: another issue we'll need to tackle is that i-PI expects structures that have a unit cell with u1//x, u2//xy, u3 generic, so an ASE parser will have to rotate.

This should be taken care of by the use of `niggliReduce` in the parser already, but I'll double check.

>  Ideally, I'd like to have a similar arrangement here: defaults to Angstrom, but if one wants to use the ase reader to parse an xyz that is in nanometers, why not?

The danger here is that depending on some formats, ASE might already be performing conversions. So if normally it would convert a format that by default is in nanometres to Angstrom, now it's going to convert another unit to nonsense. That really has more to do with the formats and their definition though, so I guess it's the kind of thing that may be left to the user. There are cases where it could make sense. By the way, any clues what the `TypeError` above is about?

ceriottm:
actually, even removing the type conversion strings, it doesn't load the file in my example.
why don't you commit an example folder with the files you are using, so we are sure to be looking at the same issue?

stur86:
> actually, even removing the type conversion strings, it doesn't load the file in my example. why don't you commit an example folder with the files you are using, so we are sure to be looking at the same issue?

It seems to work for me as long as I change the line to:

```xml
<file mode='ase'> ice_96_ase.xyz </file>
```

Or at least, it loads fine (it then seems to keep iterating something that prints ` # Interpreting input with dimension length, units angstrom and cell units angstrom`, not sure if this is intended behaviour or a different issue). Are you sure you have an up to date version of `ase`? It could be the source of the issue. I'm using `3.23.0b1` straight from their repository, though I think the latest version on `pip` should be absolutely fine too.

ceriottm:
I confirm the problem I had was from an outdated version of ASE. Now, the endless loop comes from the fact that `initializer.py/init_file:134-144 allows to read multiple frames from the init file, e.g. to initialize a full path from a multi-frame xyz. 
Seems however that somehow ase.io.read seeks back after going through the file, so it keeps reading on. IDK if my diagnosis is correct, and if this can be tackled without upending the i-PI I/O logic. glad to hear your opinion

stur86:
Ah, I see. I think the problem may be subtler than this - perhaps connected to `.xyz` files in particular, since they can contain several snapshots. I tried this locally with a different filetype and `ase` and it doesn't happen (not within `i-pi`, I just tried it on its own). If I do

```py
f = open('myfile.cif')
print(io.read(f))
print(f.readlines())
```

I get an ASE Atoms object and an empty list, while on the other hand

```py
f = open('myfile.cif')
print(f.readlines())
print(io.read(f))
```

returns a list of lines followed by an ASE error. So there's something more going on here. I'll look into it.

ceriottm:
Yes. We could force the ASE reader to throw a EOFError() after the first frame - reading multiple images from a single file is kind of ill-standardized anyway. There are also other issues related to handling of non-standard data - for instance, many file formats would store also velocities, which could make it possible to initialize both at once. 
In general with i-PI we always tried to minimize reading too much into the ontologies of other codes. 
One thing we need to pay attention to here is also that `<file>` is a bit special, as it tries to initialize at once positions, cell, and atom labels. most file formats contain all of those, but this may not be universally true for all formats read by ASE. I think we need a pragmatic solution, but one that does not run the risk of getting mysterious errors from the perspective of a user.

stur86:
So about the standardisation, most file formats will return you cell and positions - for labels, they may just not be part of the format at all, but they can be defined conventionally (e.g. we just name the atoms `H_1`, `H_2`, ... `C_1`, ... and so on). The cell may not be present in the case of non-periodic systems (like regular `.xyz` files). If you think that's something that shouldn't be supported, it's easy to add a check:

```py
if not all(a.get_pbc()):
      raise RuntimeError('File is not a periodic structure')
```

For the frame thing, here's probably the source of our issues:

```py
    # scan through file to find where the frames start
    try:
        fileobj.seek(0)
    except UnsupportedOperation:
        fileobj = StringIO(fileobj.read())
        fileobj.seek(0)
```

which is part of `ase/io/extxyz.py`. So it's a format specific problem. I guess we could fix it by adding a crude check like this:

```py
if filedesc.tell() > 0:
    raise EOFError()

try:
    a = read(filedesc)
except ValueError:
     raise EOFError()
```

ceriottm:
I tried to use tell() to see what was happening, but was getting an error. Feel free to experiment.

stur86:
Yes, it must have been the `OSError` I experience too and which I'm using the catch-all clause to get (apparently on some systems it's an `IOError`). 

I just pushed a change that should fix that, give it a try. Do we want to guard against non-periodic systems too?

ceriottm:
So this is far less trivial that it seems/seemed - mostly because both i-PI and ASE make lots of assumptions.
I pushed a fix that converts masses in a.u., because those do not get converted. Actually other i-o modules get them from the labels, so there's a bit of a confusion risk here.
I suggest that for the moment we restrict ASE reader to work with positions/cell/labels/masses so we don't have to worry about the possibiltiy it may also read a format that store velocities.
Next step would be for you to prepare a folder that serves as example/test, e.g. one in which you have a few formats and read them from the input. I would make sure to pick a non-orthogonal cell so we can see if there are issues in the conversion. 
From that POV, I don't like that niggli_reduce changes the cell also if it's orthorhombic. it should really act only if there is a mismatch with the expected i-PI format.

stur86:
Agree on the velocities (not even sure what formats ASE supports that do that anyway). I'll look up a few possible examples for the folder, how do you want that to be inserted within the structure of the repository (actually how to design tests in i-Pi was one of the things I asked originally because I can't figure it out or find documentation for it)?

About `niggli_reduce`, it's a conventional transformation meant to reduce the cell to a "standard" format, but I guess it means a lot gets changed. Should we try to just have a rotation that maps `a` onto `x`, then `b` into the `xy` plane while keeping `c` in the upper `z` semispace? That should be doable if a bit more headache inducing to do.

stur86:
Hi, so here's an example:

[triclinic_examples.zip](https://github.com/i-pi/i-pi/files/7562321/triclinic_examples.zip)

I took triclinic K2S2O8 and randomly rotated its cell to go against i-Pi's conventions. I also fixed the code so that now it should only bring it back to the conventional form (a parallel to x and b in the xy plane) without any other changes.

ceriottm:
Seems to work. I added an example based on your K2S2O8 data testing it by actually running a calculation might be good

ceriottm:
can you use another example other than a CASTEP file? it triggers some warnings that are detected as errors by the CI, because CASTEP is not present so it can't validate the input

stur86:
> can you use another example other than a CASTEP file? it triggers some warnings that are detected as errors by the CI, because CASTEP is not present so it can't validate the input

Sure, but aren't there two other versions already? ASE supports a lot of formats, any of those works. Plus ASE also converts from one format to another with this command line tool:

```
ase convert <old file> <new file>
```

so you can use that to just convert `K2S2O8_rotated.cif` to any other extension supported by ASE easily (for example `.pdb`). Here's a full list, though some of these may be designed for molecules and not allow unit cells:

https://wiki.fysik.dtu.dk/ase/ase/io/io.html

ceriottm:
ok I made this work with the CI was a bit of a pain to figure out what was the problem. Overall it looks good to me but it might be a good idea to test it some more. I'll ask other reviewers.

requirements.txt
```diff
@@ -9,3 +9,4 @@ pytest-cov
 mock
 coverage
 flake8>=3.8.4
+ase
```
litman90, line 12: Do we really want to put ase as a requirement?
I don't think so.
But if we do it, I would specify a version.

ipi/utils/io/backends/io_ase.py
```diff
@@ -0,0 +1,100 @@
...
+
+try:
+    import ase
+except ImportError:
+    ase = None
+
+__all__ = ["print_ase", "read_ase"]
+
+
+def _asecheck():
```
litman90, line 22: can we check for an ase version here as well?

tools/py/tests.py
```diff
@@ -58,7 +58,9 @@
     args = parser.parse_args()
     tests = args.tests
     print(test_folder[args.tests])
-    exit_code = pytest.main(["--tb=line", "-ra", str(test_folder[args.tests])])
+    exit_code = pytest.main(
+        ["--tb=line", "-ra", "-vv", "-s", str(test_folder[args.tests])]
```
litman90, line 62: what are these new flags for ?

litman90 (commented):
I am not a regular use of ase.
But when I use it 
I normally specify the format when reading. something like
read(name, format="aims")

I think to exploit the full potential  of the ase's parser, thinking about formats and not quantities, we need to pass the 'format' key 


A small detail for @ceriottm, after running the examples I noted that the kinetic energy is not the same for both outputs. I guess that for the second run the thermostat is not starting from the same state.

requirements.txt
stur86: Do you have a specific case in mind where specifying the format is important? In my experience, it's mostly irrelevant - ASE internally checks the file name and deduces it from the extension, which in virtually all cases will be enough.

stur86: (sorry, this comment of course referred to your first observation about the use of the method `read`, not the version/dependency thing)

litman90:
Other than that, the rotated cell test is in principle the 'trickiest' test, and it looks fine.

requirements.txt
ceriottm: I believe it was me adding ASE to the requirements - problem is that if we don't the CI fails. if you can make it a requirement only for the CI, then I'm happy to drop it here, but the test will fail when launched locally if ASE is not there

tools/py/tests.py
ceriottm: to make sure that when the regtests fail, they will tell you why.

requirements.txt
stur86: Can't we make the test itself optional? Just skips it altogether if an attempt to load ASE gives an `ImportError`.

ceriottm: that would be ideal but I think we don't have the infrastructure for that. @litman90 might be in a better position to tweak the tests infrastructure in this direction.

litman90:
I took out the ase from requirement.txt and added to github actions infrastructure.

litman90:
I moved the units section a level-up inside the github webpage, and also added a warning regarding using ASE.

"
 When using ASE format for I/O, the units are dictated by the ase defaults.
    So for example, the ASE extended xyz format expects positions in angstroms.
    (See for example ``examples/ase-io/``)
"

stur86:
Hi @litman90 @ceriottm  , is there anything else left to do here? I am going to leave soon and might not be able to work on this any more, so I'd like to get the PR done by next week. Thanks!

litman90:
For me, it is good to be merged. @ceriottm can merge it if he agrees.

ceriottm:
Looks good to me - and seems at the very least it breaks nothing. We can iron out potential issues later on.

PR #197, opened 2021-09-06
Fix bug and update sGDML force field function

sauceda:
There was a couple of problems getting units from sGDML models trained with the newest version of the software. Additionally, some troubles were appearing when loading the model file without the key "allow_pickle=True".


ceriottm:
Sorry for the lag @sauceda , shall we merge this? Seems small and to the point so if you just run the linter I expect this will be a fast one.

venkatkapil24:
@sauceda all you need to do is run i-pi-style in the base directory (i-pi/), and it should be okay for a merge. Let me know if you need help.

venkatkapil24:
@sauceda sorry for not being clear but there is an executable called i-pi-style in bin/ which makes sure that the indentation and style are consistent with the rest of the code. Can you run that and then we can merge it ASAP. Feel free to reach out if it is not clear.

venkatkapil24:
Hey @sauceda if you can run the requested operation by tomorrow (should take 5s) I can merge this. Otherwise I will have to close this one.

PR #199, opened 2021-12-06
Increased precision in XML to %18.10e

fidanyan:
As we discussed, I increased precision of floats by 2 digits. I decided not to reduce spacing between entries, two spaces are nice and we probably can survive with that. This is the dummiest fix possible, but since it solves the problem, I think it's the way to go now. If we agree on something nicer later, we can implement it, but right now it's just needed to do calculations.

### DIFF
--- a/ipi/inputs/simulation.py
+++ b/ipi/inputs/simulation.py
@@ -22,8 +22,11 @@ import ipi.inputs.outputs as ioutputs
 from ipi.inputs.smotion import InputSmotion
 
 
-__all__ = ["InputSimulation"]
+__all__ = ["InputSimulation", "GLOBALS"]
 
+# Apparently, the easiest way to propagate format to all output cases
+# is to define it globally.
+GLOBALS = {"floatformat": "%18.10e"}
 
 class InputSimulation(Input):
 
@@ -124,6 +127,14 @@ class InputSimulation(Input):
                 "options": ["md", "paratemp", "static"],
             },
         ),
+        "floatformat": (
+            InputAttribute,
+            {
+                "dtype": str,
+                "default": "%18.10e",
+                "help": "A format for all printed floats.",
+            },
+        ),
     }
 
     dynamic = {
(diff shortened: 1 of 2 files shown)


ceriottm:
Is adding an option tricky? Would be nice to try actually, it's more sustainable long-term and in many cases one may want to *cut* a few digits to save space

ceriottm:
I was thinking of something even uglier: a global!

fidanyan:
Hm, global would be way easier. 
I try to let it through the pipeline, and it goes very deep. For example, XML output binds to Simulation, while others bind to system(s). So, this new attribute will have to appear literally everywhere.

ceriottm:
If we go down the GLOBALS way, you can take `softexit` as an example. I would add a `_floatformat = '%18.10e"` to utils/io/__init__.py, then you can import it and should be able to set it.

fidanyan:
Ah, sorry, didn't see your comment)

fidanyan:
> If we go down the GLOBALS way, you can take `softexit` as an example. I would add a `_floatformat = '%18.10e"` to utils/io/**init**.py, then you can import it and should be able to set it.

I tried with a variable first, but didn't figured out all these mutable/immutable issues. I guess instance of a class behaves more like a container rather than like a variable. Containers behave more predictably, and a dict looks better than a list if we at some point decide to add more properties. I can move it to `utils/io/__init__.py`, but since I have to fetch it anyways from simulation, I thought that `inputs/simulation.py` it's a right place.

fidanyan:
Michele, do you think GLOBALS _should be_ in `utils/io/__init__.py`? If yes, I will create a clean branch and do it there. If it's okay as present (in `inputs/simulation.py`), then I think it's done. But I'd like to make a clean branch and redo it there, because the commit history becomes unnecessarily dirty for a couple of lines - unless you can squash upon merging (can you?).

ceriottm:
if we should have an overal dict of globals, I would have it as part of the
main i-PI __init__

ceriottm:
so you can do say `ipi._global_settings['float_format']`

fidanyan:
See #201 for clean implementation

PR #200, opened 2021-12-08
Fixing a bug where atomswap fixes would not work with REMD

ceriottm:
It attempts to swap the thermostats, but not all motion classes have a thermostat or a barostat.


ceriottm:
This is harmless enough - doesn't touch any important file - and is a bugfix, so I will just merge it

PR #201, opened 2021-12-10
Added GLOBALS to ipi/__init__.py, made default float %18.10e

fidanyan:
Here is a clean single-commit implementation with GLOBALS in `ipi/___init__.py`

### DIFF
--- a/ipi/inputs/simulation.py
+++ b/ipi/inputs/simulation.py
@@ -129,7 +129,7 @@ class InputSimulation(Input):
             InputAttribute,
             {
                 "dtype": str,
-                "default": "%18.10e",
+                "default": "%16.8e",
                 "help": "A format for all printed floats.",
             },
         ),
--- a/ipi/inputs/simulation.py
+++ b/ipi/inputs/simulation.py
@@ -5,7 +5,7 @@
 # See the "licenses" directory for full license information.
 
 
-from ipi import GLOBALS
+from ipi import ipi_global_settings
 from ipi.utils.depend import *
 from ipi.utils.inputvalue import *
 from ipi.utils.units import *
@@ -191,7 +191,7 @@ class InputSimulation(Input):
         self.total_time.store(simul.ttime)
         self.smotion.store(simul.smotion)
         self.threading.store(simul.threading)
-        self.floatformat.store(GLOBALS["floatformat"])
+        self.floatformat.store(ipi_global_settings["floatformat"])
 
         # this we pick from the messages class. kind of a "global" but it seems to
         # be the best way to pass around the (global) information on the level of output.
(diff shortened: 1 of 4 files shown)


ceriottm (commented):
I'd call this something like `ipi_global_settings` rather than `GLOBALS` (python has its own __globals__ variable I think) but otherwise, as ugly as it is, I guess it's the best quick-fix for this problem

fidanyan:
Renamed

ceriottm:
are there any examples?

ipi/inputs/simulation.py
```diff
@@ -273,9 +291,10 @@ def fetch(self):
             if k == "system":
                 syslist.append(v.fetch())
             elif k == "system_template":
-                syslist += (
```
ceriottm, line 276: Why this change? I think the multi-line comment was more readable

ipi/utils/io/inputs/io_xml.py
```diff
@@ -533,10 +533,10 @@ def write_tuple(data):
 def write_float(data):
     """Writes a formatted string from a float.
 
-    Floats are printed out in exponential format, to 8 decimal places and
-    filling up any spaces under 16 not used with spaces.
+    The default format is exponential to 10 decimal places and
+    filling up any spaces under 18 not used with spaces.
 
-    For example 1.0 --> '  1.00000000e+00'
+    For example 1.0 --> '  1.0000000000e+00'
```
ceriottm, line 539: I would do the % 17.9e format so it's space-neutral WRT the current format

ceriottm (changes requested):
I have only those two small comments, and then it'd be good to have this in at least one of the examples, so it's tested and there is an example of using it.

ipi/inputs/simulation.py
fidanyan: Not sure what you mean. I noticed that Black exploded a simple expression over many lines because of long comment on the same line, so I moved the comment above the expression and split manually. Mainly to prevent exploding the expression.

ipi/utils/io/inputs/io_xml.py
fidanyan: Not sure what you mean. 17.9 isn't neutral with current 16.8 anyways. I guess you mean 16.9 with sacrifice of one whitespace? But worth noticing that there can be a minus sign which takes place of one of two whitespaces. I wouldn't leave only one space because of that.

ceriottm:
ok I misunderstood the diff. I like how you changed it.

ceriottm:
yeah I meant 16.9 with one space separator

fidanyan:
My comment about minus sign holds. But, in general, I feel that you don't like the idea of adding digits into default. TBH, since *.9 is anyways not enough to have meV precision in my system, I don't mind to leave the default 16.8 as it was before - just to stop wasting our time discussing this. And I'll just set 18.10 in my calculations.

Regarding an example - I'll add `floatformat` to one of the fhi-aims examples.

ceriottm (approved):

PR #202, opened 2022-01-25
patch for VASP 6.2.1

KaiSellschopp:
I discovered that the patch for VASP 5.4.4 does not work with VASP 6.2.1 because they introduced the new function PROCESS_INCAR in src/reader.F, which caused the host and port to always default to localhost:23333. I created an adapted patch for VASP 6.2.1, which should also work for other versions of VASP that make use of PROCESS_INCAR for reading the INCAR tags. Apart from the changes to reader.F the patch is basically the same as for 5.4.4, just some line numbers and surroundings of the code additions changed.


ceriottm:
Actually there is a long-standing PR at VASP to merge the i-PI interface in
the mainline. So drop them a line, they might finally decide to go ahead
and do the merge.

KaiSellschopp:
I was wondering if they would do that, especially since the work is basically done already and they only need to put it in the source code. But as always, it takes a bit of time until features appear in the official releases, I guess. I'll write them to give some extra motivation ;)

KaiSellschopp:
There was a new release now (6.3.0) that didn't include the i-PI interface. I also added a patch for this new release to this pull request, and dropped them a message for future releases. Would be great not having to patch each version individually.

alchem0x2A:
Hi, @KaiSellschopp @ceriottm thanks for your PR with the iPI interface integrated with VASP 6! I'm currently developing a socket-client calculator interface for VASP based on its built-in "interactive mode" (https://github.com/ulissigroup/vasp-interactive#using-vaspinteractive-as-socket-client-calculator-driver-mode). 

For reference purposes, we also build VASP binaries (5.4 ~ 6.3) with iPI socket patched following this PR, using a version-independent patch script (https://github.com/ulissigroup/vasp-interactive/blob/main/vasp-build/patch_ipi.py). Since the script uses regex patterns to replace source code blocks, it may survive future VASP releases if iPI is not yet in VASP by then. I would like to know if you think it makes sense to include such script in the VASP examples. Thank you!

KaiSellschopp:
Hi @alchem0x2A, I think that's a great idea that would most likely reduce the effort for future VASP versions. The interactive mode is news to me, and I can't find anything about it in the official docs. Is that a new feature? Maybe they are implementing this interface as a generalized one to not have to build interfaces with individual codes?
Just out of curiosity: did you also try to make the patch work with the openACC/openMP-variant to be able to run on GPUs? I tried, but failed so far.

venkatkapil24:
> 

@alchem0x2A I am happy to help you add an example. You can add it here and I will review the PR.

venkatkapil24:
This can be closed. PR #250  will take care of this.

PR #203, opened 2022-02-15
Bugfix in trimsim.py and getacf.py

fidanyan:
I noticed a couple of small bugs it `tools/py/`. Trimsim was complaining that `'Smotion' object has no attribute 'swapfile'`, even though I have no smotion in my simulation. I just put `if hasattr()` for that attribute and skip that part if not. Works fine for me.

In `getacf.py`, I noticed that species names consisting of more than one character behave weird, so I added a `split()`, so that if one provides species separated by whitespaces, the script reads them correctly.


fidanyan:
As you can see, Black goes crazy..

mahrossi:
Can you fix Black here too, as this is then easy to merge? Thanks!

fidanyan:
> Can you fix Black here too, as this is then easy to merge? Thanks!

Here you go!

Makefile
```diff
@@ -1,6 +1,6 @@
 .PHONY: lint, pretty
 lint: 
-	flake8 --select=F --ignore= --ignore=F403,F405 --per-file-ignores=**__init__.py:F401 ./
+	flake8 --select=F --ignore= --ignore=F403,F405 --per-file-ignores=**__init__.py:F401 ./ --exclude=my_tests
```
mahrossi, line 3: What are you excluding here, and why?

mahrossi (changes requested):
Overall ok, just an exclusion added to Makefile that is unclear. Just clarify and then it should be fine.

Makefile
fidanyan: Ahh, sorry, I shouldn't have committed the Makefile. It's just my local folder with tests that I store inside the repo. I'll change this back.

mahrossi (approved):

mahrossi (approved):

PR #204, opened 2022-02-28
Lint/black update

ceriottm:
I believe that black has updated its style guidelines. I updated locally to check if this fixes the linting problems.


ceriottm:
Can we merge this?

litman90 (approved):

Issue #206, opened 2022-05-05
Negative pressure supports for barostats


salinelake:
It will be useful to allow the simulation of anisotropic NPT ensemble with a negative pressure or stress. Sometimes people are interested in study phase diagram under these conditions.

ceriottm:
Doesn't it work already? There should be no particular difficulty in running with negative pressure, as long as the system can withstand the tension without breaking apart

salinelake:
> Doesn't it work already? There should be no particular difficulty in running with negative pressure, as long as the system can withstand the tension without breaking apart

Good to hear. Then I guess the only issue is the value check needed to removed in engine/motion/dynamics.py.
https://github.com/i-pi/i-pi/blob/d95d011f32938629e0388534225eced4b71a6900/ipi/engine/motion/dynamics.py#L223

ceriottm:
ok cool. this is not the first time someone asks, and we had suggested to just hijack those ifs, but given this is more common than we though can you check if #207 works for you? hope you don't need to run at -12345 atomic units of pressure...

salinelake:
> ok cool. this is not the first time someone asks, and we had suggested to just hijack those ifs, but given this is more common than we though can you check if #207 works for you? hope you don't need to run at -12345 atomic units of pressure...

Thanks, it works!

PR #207, opened 2022-05-10
Allow negative pressures

ceriottm:
Mini-fix to allow deliberate negative pressure in npt and nst simulations


ceriottm:
Fixes #206

PR #209, opened 2022-06-01
Add a check for the type of data in driver.py

Luthaf:
We had an issue where the forces where returned as a 32-bit floating points, which then crashed the i-pi side which expected to get 64-bit floating points. This PR adds an explicit check for this in the driver code.


drivers/py/driver.py
```diff
@@ -122,6 +122,18 @@ def run_driver(unix=False, address="", port=12345, driver=Dummy_driver()):
...
             sock.sendall(Message("FORCEREADY"))
 
+            if not isinstance(forces, np.ndarray) and forces.dtype == np.float64:
+                raise ValueError(
+                    "driver returned forces with the wrong type: we need a "
+                    "numpy.ndarray containing 64-bit floating points values"
+                )
+
+            if not isinstance(vir, np.ndarray) and vir.dtype == np.float64:
+                raise ValueError(
```
Luthaf, line 132: This could also check for the matrix shape, if the data is contiguous (not sure what the behaviour of send_data would be with non contiguous arrays), and if it is stored in C order.

ceriottm (approved):

PR #223, opened 2022-07-28
Added CASTEP example

mijp1:
CASTEP v22 onwards has nature support for sockets and hence can be used with i-pi.
This is a trimmed down example of a 4-bead PIMD run using i-pi and CASTEP, with a simple script to automate launch & execution.


mijp1:
Note: this PR does not change any of the i-pi source - it is just an addition to the "examples" showing how CASTEP v22 onwards now works natively with i-pi

examples/castep/run_me
```diff
@@ -0,0 +1,71 @@
...
+
+#all present so now all OK to launch i-pi
+set nbeads = `grep nbead ${seed}.xml | awk -F\' '{print $2}'`
+if ( "$nbeads" == "" ) then
+   set nbeads = `grep nbead ${seed}.xml | awk -F\" '{print $2}'`
+endif
+
+echo Starting i-pi with $nbeads beads for system $seed
+i-pi ${seed}.xml > log &
+
```
litman90, line 56: I would add a  'sleep 5' here since in some machines i-pi takes a  few seconds to initialize.

litman90 (changes requested):

litman90:
Thanks for the contribution!
I activated the automatic test which should pass without problems. Please add the line I suggested. Afterwards, we are ready to merge

ceriottm:
thanks @mijp1 for the contribution, and thanks Yair for vetting it. I'll
also add a reference to CASTEP among the list of codes that support the
i-PI interface.
Best
Michele

examples/castep/run_me
mijp1: OK - now done.
    Matt

litman90 (approved):

PR #224, opened 2022-08-05
Fixes and polish of the constrained MD code

ceriottm:
This is to give some love to the constraint code, that has been largely unused and became plain broken during the py2->py3 conversion. This PR is to coordinate fixing the conversion bugs, and making sure the code is usable and reasonably trustworthy.


nickgerrits:
My apologies for the late reply. I've been very busy lately and the coming time so input from my side will be somewhat limited for now.

I think what needs to be changed is the way the constrained is applied. Right now it works in cartesian coordinates for the beads, while only applying the constraint on the first bead, i.e., it is only implemented for the classical case. I think by switching to normal mode coordinates for the beads and applying the constraint only to the first normal mode (i.e., the centroid), while leaving the other normal modes untouched, should fix the problem. That way you can use it both in RPMD and classical MD. From my understanding of the current code, this would also be a relatively low effort change since a lot from the regular dynamics code can be reused.

I'm not convinced that the matrices that need to solved for the constraints are guaranteed to be positive definite. Sure, Cholesky decomposition might be the fastest solution, but at the very least a fallback option needs to be offered then (for example through a try except construction).

I'm not sure yet whether coupling the number of removed DOFs with ncons is the right way to go. I would need to have another look how everything is connected (because ncons is also related to the solver).

Some additional constraints are easily added such as a fix position or com constraint. The com constraint can be improved beyond the standard one by allowing the com of specific indices to be constrained instead of the entire system. Additional addition can be constraints such as a line or plane constraints, but I consider this to be of considerable less importance than the other stuff raised so far.

I'll try to work at some of this stuff when I finally have some more time again, but I figured it would be good to start the discussion.

ceriottm:
Any appetite for moving this forward? Otherwise I'd merge at least the bugfix and just trigger an error when it's called with nbeads>1

litman90:
I am interested in this feature but don't have time at the moment to work on it. I will happily take it as my main task in a soon-to-be hackathon =)

nickgerrits:
For now it is indeed better to merge the error trigger. I don't have time right now to work on it but I might be able at some point (or perhaps help Litman).

As for what was mentioned in the other thread: I believe constraining the centroid should be ok, although it is of course not very rigorous. This would also be a relatively straightforward fix of the code, where you have a similar setup as in the regular dynamics code but with the centroid normal mode being constrained. We could also allow the other modes to be constrained as well as an option. However, in the current PIMD calculations you can also constrain individual atoms or the COM. As such, I don't see how doing this for other constraints is any different. My original goal was to perform 1D NVT calculations on vibrational modes of a molecule, for which I have also developed a constraint. But in any case, it should be possible to define any kind of constraint and it should be ok. For example, one might want to use a line or plane constraint to restrict movement of atoms to a vector or plane. This is also what is done in VTST where hyperplanes (which coincides with a plane constraint) need to be sampled, e.g. through MD simulations, to advance the hyperplane towards the free energy maximizing dividing surface (basically a collection of TSs instead of just the first order saddle point while also including temperature effects). As far as I understand, there are also path integral extensions of such methods that have been quite successful. To me intuitively, this would mean that constrained dynamics within the RPMD framework should be at least reasonable. The big question would be rather if every normal mode should be constrained or only the centroid.

I would also like to re-emphasize that even if you were to use this code only for classical simulations (i.e., Nbeads=1), the constraint solver is far from robust in terms of numerics. I feel this could be improved by implementing different checks and fallback options for the solver. Or move away completely from doing this numerically and doing it analytically instead, which is how for example ASE deals with constraints.

PR #225, opened 2022-08-06
Inconsistency in the computation of the forces and virial in the baseline + commitee model

DavideTisi:
In the baseline + committe models the variance of the committee is defined as (see also Eq. 11 in [https://doi.org/10.1063/5.0036522](doi:10.1063/5.0036522 )):

$$\sigma^2 = \frac{1}{M-1}  \sum ^M _{i=0} (V_i - \bar{V})^2$$

where the sum runs over the $M$ models of the committee, $V_i$ is the potential of $i$-th model and $\bar{V}$ the mean potential of the committee.
the formula is correctly implemented in the variables `var_pot` , where the sum is divided by $(M-1)$ .

for the derivative the term `uncertain_frc` and `uncertain_vir` contained a $1/M$ which created an inconsistency.
This PR fixes this issue.

PR done by Davide Tisi, Lorenzo Gigli e Federico Grasselli


ceriottm:
Hello, thanks for this PR. 
(1) does this fix energy conservation, for a "well-aligned" baseline?
(2) please make sure this is linted properly (instructions are in the README, make sure you have a recent version of black)

ceriottm (approved):

PR #227, opened 2022-08-12
Regtest upgrade

litman90:
Simple addition to incorporate the possibility to define one tolerance value for all columns in NumPy data.


ipi_tests/regression_tests/runstools.py
```diff
@@ -65,17 +65,25 @@ def run(self, cwd, nid):
...
-                        listll.append(int(lltol[0]))
-                        if len(lltol) > 1:
-                            listtol.append((float(lltol[1]), float(lltol[2])))
-                        else:
-                            listtol.append(None)
-                    self.usecol.append(listll)
+                    info_tol = line.split()[2].split(",")
+                    if info_tol[0] == "all":
+                        listcol = info_tol[0]
+                        listtol.append((float(info_tol[1]), float(info_tol[2])))
```
ceriottm, line 76: I think you should write it such that it also supports the case in which you have "all" without tolerances. It is a bit redundant, but it is more consistent with the structure when you can specify columns also with default tolerances

ipi_tests/regression_tests/tests/NEB/FIRE/zundel/files_to_check.txt
```diff
@@ -1,6 +1,6 @@
         filename                        format      icol,rtol,atol  icol,rtol,atol ...
 --------------------------------------------------------------------------------------
-ref_simulation.neb                      numpy       0,0.0,0.0  6,1e-7,1e-6  7,1e-7,1e-6  8,1e-7,1e-6  9,1e-7,1e-6  10,1e-7,1e-6  11,1e-7,1e-6
+ref_simulation.neb                      numpy       all,1.e-7,1e-6
```
ceriottm, line 3: I would also use the all syntax for the phonon case. There there are several checks on dynmat-like files with a load of columns, and I had only selected a random few. It's a great use case for `all`

ceriottm (changes requested):

litman90:
I  think I applied all the requested changes, let me know if something else is required to merge this PR.

ceriottm (approved):

PR #232, opened 2022-10-21
Polished the logic to make ASE output (to extxyz for starters) work.

ceriottm:
Also establish clearly that ASE (and document in manual) that ASE interprets xyz as written in angstrom. This will probably save some headaches in the future


litman90 (dismissed):
Everything looks fine.
I will add an issue as a reminder for ourselves to add a regtest for i/o with ase when we have time.
If you want to address it now great, but it's not necessary to approve this PR

PR #238, opened 2023-01-19
Update README.md

ianhbell:
The folders must have moved, README out of date.

PR #239, opened 2023-01-19
update of workflow

litman90:


litman90:
Let me try closing and re-opening to see if the new tests can be triggered

litman90:
Well, we can set up conda now, and the tests are triggered.
However, many tests  'Could not find the i-PI UNIX socket.' but some can.  
I don't see a clear pattern of success/failure. I can't check it further now, will do it in the next few days.

litman90:
In short, we had some deprecation problems with NumPy and GitHub-CI. @ceriottm Please have a look, approve and merge. This should fix the problem for #238

ceriottm (approved):

PR #240, opened 2023-01-25
This is just the modification of a post-processing script

mahrossi:
We would like to modify this post-processing script for NEB. It was pushed directly to master in our repo.
In our repo, the workflows are broken, and I could not fix them. It has nothing to do with the actual regression tests, just conda version (it seems). Hating github ATM :-D

### DIFF
--- a/tools/py/neb_interpolate.py
+++ b/tools/py/neb_interpolate.py
@@ -11,6 +11,8 @@ Arguments:
     --units             - i-pi-supported distance units to write output geometries.
     -al, --activelist   - whitespace-separated list of active atoms to read from the command line
     -af, --activefile   - a file with the space or linebreak separated list of active atoms.
+    -o, --output        - Name of the output xyz file (including extension). Default is "interpolated_path.xyz".
+    --dry               - a boolean flag to switch off output into file(s).
     EITHER
       --ini         - initial geometry (.xyz) if mode "endoints"
       --fin         - final geometry (.xyz) if mode "endpoints"
@@ -28,9 +30,11 @@ import os
 import numpy as np
 import sys
 import argparse
+from numpy.linalg import norm as npnorm
 
 from ipi.utils.io import read_file, read_file_raw, print_file
 from ipi.utils.io.io_units import auto_units
+from ipi.utils.units import Constants
 
 try:
     import scipy
@@ -52,7 +56,6 @@ def spline_resample(q, nbeads_old, nbeads_new, k=3):
     Returns:
         new_q - resampled coordinates
     """
-    from numpy.linalg import norm as npnorm
 
     if scipy is None:
         print("spline interpolation requires scipy module.")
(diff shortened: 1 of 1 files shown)


ceriottm:
See #239 might be best to first merge that.

ceriottm:
I think you need to create a branch and open a PR from there, or merge the current upstream master from the CLI, as it does not let me merge upstream master from here

venkatkapil24:
@mahrossi is this ready? if you think this is giving the expected output, I am happy to merge.

tools/py/neb_interpolate.py
```diff
@@ -276,6 +287,12 @@ def spline_resample(q, nbeads_old, nbeads_new, k=3):
             print("Error: cannot find {}.".format(input_fin))
             exit(-1)
 
+        if np.all(inimasses == finmasses):
+            masses = inimasses
+        else:
+            print("Error: initial and final masses differ.")
```
venkatkapil24, line 293: Could the error be made more transparent for the user? Perhaps with a suggestion on what they should do. 
Does this mean the initial and the final structures (two ends of the NEB) have different orders of atoms or something?

tools/py/neb_interpolate.py
```diff
@@ -305,6 +322,12 @@ def spline_resample(q, nbeads_old, nbeads_new, k=3):
                     exit(-1)
 
                 natoms = path[0].natoms
+                masses = path[0].m
+                # Check that the masses in all frames are the same
+                for atoms in path:
+                    if np.any(atoms.m != masses):
+                        print("Error: masses differ across XYZ frames")
```
venkatkapil24, line 329: If a suggestion on what they should do can be added, it could help the user. Like, make sure...

tools/py/neb_interpolate.py
```diff
@@ -344,6 +368,9 @@ def spline_resample(q, nbeads_old, nbeads_new, k=3):
     # Below this point, it shouldn't matter where the inputs came from.
     # =================================================================
     print("Imported structures: %d geometries, %d atoms each." % (nbeads_old, natoms))
```
venkatkapil24, line 370: Could the print statements be consistent with the i-pi output like "@NEB: Importer structure"

tools/py/neb_interpolate.py
```diff
@@ -344,6 +368,9 @@ def spline_resample(q, nbeads_old, nbeads_new, k=3):
...
+    print("Atomic masses:")
```
venkatkapil24, line 371: print formatting

venkatkapil24 (commented):
(1) The print statement can be consistent with the other motion classes with an @NEB-like prefix to all the statements. (2) The error messages can be more helpful with a suggestion on what the user could potentially change.

mahrossi:
Hey @fidanyan - if you are still around can you do these minor tweaks?

mahrossi:
Hey @venkatkapil24 - I think @fidanyan is not going to answer this very soon. I am out of bandwidth for the next 3 weeks. This is a post-processing script and all comments, even though they would improve the user experience, are not critical. Can this be merged?

venkatkapil24 (approved):
Yeah in that case we can. Although it would be good if someone can clean it up in future.

tools/py/neb_interpolate.py
fidanyan: Hi Venkat! I'm not sure about the @NEB prefix, because it's a standalone utility which is not meant to be called from the main code. I think it is a reasonable suggestion in the `spline_resample()` because it is designed to be reusable, but not in `__main__()`.

PR #241, opened 2023-02-05
Removed virial from harm3d potential in driver.f90. 

BarakHirshberg:
Hi guys, 
Working on a new (much faster) implementation for bosonic PIMD, in preparation for our tutorial in the CECAM school.
Along the way, made a minor change here. I deleted the virial from harm3d potential.
It is not mandatory and had invalid memory access for more than three particles.
Barak

### DIFF
--- a/drivers/f90/driver.f90
+++ b/drivers/f90/driver.f90
@@ -472,15 +472,11 @@
             ELSEIF (vstyle == 30) THEN ! 3D harmonic potential
                    pot = 0.0d0
                    forces = 0.0d0
-                   virial = 0.0d0
                DO i=1,nat 
                    pot = pot + 0.5*ks*atoms(i,1)**2 + 0.5*ks*atoms(i,2)**2 + 0.5*ks*atoms(i,3)**2
                    forces(i,1) = -ks*atoms(i,1)
                    forces(i,2) = -ks*atoms(i,2)
                    forces(i,3) = -ks*atoms(i,3)
-                   virial(i,1) = forces(i,1)*atoms(i,1)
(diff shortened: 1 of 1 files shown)


venkatkapil24:
Hey Barak, harm3d is used in many examples to do dummy npt. Perhaps removing that triggered the example error.  Any chance we can keep harm3d virial? Else @litman90 was there a particular reason to use it other than say gas?

BarakHirshberg:
I don't mind keeping it at all. It was just implemented wrongly, probably my fault to begin with.
I think we can easily implement the correct virial and then we can keep it.
Do you want me to look into it?  I am flying to CA tomorrow but can probably get to it next week.

venkatkapil24:
Yeah that would be the best option.

BarakHirshberg:
Hi Venkat and Michele,

I implemented the virial for the harm3d.
I wasn't sure what is the current index definition, so could you please check?
Couldn't find it anywhere.
If just the diagonal elements are needed, this should matter, but I wasn't sure if that's the case.
Also, I saw somewhere else that a small nonzero value is given to the virial so the NPT runs don't fail.
We could also do that if the only goal is to make the tests run.

Thanks,
Barak

mahrossi:
Hey @BarakHirshberg, I had a check on the definition in `i-pi/drivers/f90/LJ.f90` and I think you defined it in the switched order (you can have a look there too to confirm). Can you switch to v(i,j) = f(i)*atoms(j)? then we should be able to merge.

BarakHirshberg:
Done! :)

Issue #221, opened 2022-07-26
Constrained dynamics


nickgerrits:
Dear all,

I've been working on constrained dynamics with i-Pi and I've noticed some peculiar things and bug(s).

I've tried running some test simulations involving the constraints "distance" and "angle" but they don't seem to work. This seems to be a bug in inputs/motion/constrained_dynamics.py:
135: "alist.shape = (alist.shape[0] / 2, 2)"
145: "alist.shape = (alist.shape[0] / 3, 3)"

Changing this to
135: "alist.shape = (int(alist.shape[0] / 2), 2)"
145: "alist.shape = (int(alist.shape[0] / 3), 3)"
seems to resolve the problem. I'm not sure whether this is caused by the use of particular Python versions, but in my case this worked.

Another more worrying thing is the fact that within constrained dynamics the atomic positions within each bead don't seem to be updated, only the centroid and first bead. For example, when dumping the atomic position of an atom within each individual bead through printing atom_x or atom_x_path in properties shows clearly that only the positions of the centroid and first bead are changing. Also, I've attached LAMMPS as a driver and let it dump all positions so I could check whether positions were actually changing or not. And indeed, only the position of the first bead changes, whereas the positions of all the other beads remain constant.

Weirdly enough, the momenta are adjusted for all beads (e.g., use atom_v in properties) but I'm not sure whether they are correct. I've also implemented my own constraint where atomic positions can be fixed, which works somewhat similar to fixatoms. (Note that fixatoms and fixcom are not implemented within constrained dynamics, hence the need for a fixatom constraint.) When you apply such a fix you would expect momenta to be close to zero (since you are solving the constraint numerically, you might deviate from zero due to numerical noise), but I definitely see much larger values than that, even when positions are not changing.

Either I'm really confused on what constrained dynamics is trying to achieve, or something is really going wrong. In its current form it basically only updates the positions of a single bead and it seems to not touch the positions of the other beads. This also results in the constrained dynamics to yield something completely different compared to classical (i.e., single bead) simulations: The centroid is an average of the static non-updated beads and the single bead that is correctly (?) following the constrained dynamics.

[...]

PR #246, opened 2023-02-16
Clean up and improve constraints code

ceriottm:
This is a PR to complete the work from #224 , that was merged to address the most urgent bugfixes. 
Constraints are implemented but half-working and need some love to be fully usable.


ceriottm:
See also #221 and tagging @nickgerrits and @MatthiasSachs

mahrossi:
This looks like a pretty clean and clear PR to me. 

Are there regtests for the constrained dyn? I could not find them. Wouldn't it be good to add them here?

MatthiasSachs:
Apologies for remaining silent throughout the whole process so far. 

One (very late) comment regarding the occasional occurrence of singular Gramians, which @nickgerrits raised. I agree that having additional safeguards to deal with singular Gramians would be good. Having said that, I also briefly discussed this with Ben Leimkuhler at some point and he mentioned [this article](https://onlinelibrary.wiley.com/doi/epdf/10.1002/jcc.540161003?saml_referrer=), according to which degenerated Gramian only occur in the presence of "dependent constraints", e.g., a collection of distance/bond constraints that form a ring. 

I will be a bit tied up with teaching in the weeks to come, but if there are things that would need to be urgently done, please let me know!

mahrossi:
Hi, is this ready? It does not seem like previous comments were addressed.

PR #247, opened 2023-02-25
Adding the cavity molecular dynamics functionality and example

TaoELi:
Hi Yair,

I have added the CavMD feature for simulating vibrational strong coupling to the i-pi code. 

Please take a look.

Best,

Tao


litman90:
Great, thank you Tao!
I will have a look at this PR in the next these days and will set the regression test with the data you sent me 
Best,
Yair

examples/lammps/h2o-cavmd/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+In this simulation, an infrared cavity mode at 3550 cm<sup>-1</sup> is strongly coupled to the O-H stretch mode of liquid water along both the x and y directions. Due to this strong coupling, the O-H stretch band of liquid water is splitted to two eigenstates, known as polaritons.
+
+See the paper [Proc. Natl. Acad. Sci., 2020, 117 (31), 18324–18331](https://doi.org/10.1073/pnas.2009272117) or [https://github.com/TaoELi/cavity-md-ipi](https://github.com/TaoELi/cavity-md-ipi) for more details of CavMD.
+
+This example is the same as the [Rabi splitting tutorial](https://github.com/TaoELi/cavity-md-ipi/tree/master/tutorials/Rabi_splitting) in [https://github.com/TaoELi/cavity-md-ipi](https://github.com/TaoELi/cavity-md-ipi). The only difference is that in the original [Rabi splitting tutorial](https://github.com/TaoELi/cavity-md-ipi/tree/master/tutorials/Rabi_splitting), the parameters for the photons are controlled by a separate input file **photon_params.json**. Here, instead, the photon parameters are controlled in the **ffcavphsocket** section of the i-pi input file **input.xml**. 
+
```
litman90, line 11: I would suggest adding something along the lines:

"Important: At the moment this implementation assumes that the total dipole moment is a sum of atomic charges which are fixed during the simulation while an extension to use on-the-fly dipole moments obtained from the drive is possible, it has not been implemented at the moment. If you are interested in this feature please contact us.

examples/lammps/h2o-cavmd/README.md
```diff
@@ -0,0 +1,39 @@
...
+In short, to perform CavMD simulations, we need to make **two modifications** of conventional i-pi simulations.
+
+1. We use the following force evaluator **ffcavphsocket** instead of the original **ffsocket**. In this new force evaluator, **E0** denotes the effective light-matter coupling strength in atomic units (corresponding to the \tilde{varepsilon} parameter in the [original paper](https://doi.org/10.1073/pnas.2009272117)). **omega_c_cminv** denotes the cavity photon mode frequency in wave number. **charge_array** defines the partial charges of all atoms in the same order as the input xyz file. Note that the total size of **charge_array** should match the total number of nuclei.
+
```
litman90, line 15: It would be nice and better to have a size check when reading this

examples/lammps/h2o-cavmd/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+  <ffcavphsocket name='lammps' mode='unix' pbc='True'>
+    <address>h2o-cl-cavmd</address>
+    <latency> 1e-3 </latency>
+    <apply_photon> True </apply_photon>
+    <E0> 4e-4 </E0>
+    <omega_c_cminv> 3550.0 </omega_c_cminv>
+    <charge_array> [-0.8192, 0.4096, 0.4096, ...] </charge_array>
+  </ffcavphsocket>
+```
```
litman90, line 25: Unless specified we use atomic units everywhere.
Thus, if we want to use wave number it should look like 
<omega_c_cminv units='inversecm'> 3550.0 </omega_c_cminv>
Could you do this change?


Could you modify the omega_c_cminv input to reflect this?

ipi/inputs/forcefields.py
```diff
@@ -831,3 +833,225 @@ def fetch(self):
...
+            },
+        ),
+        "timeout": (
+            InputValue,
+            {
+                "dtype": float,
+                "default": 0.0,
+                "help": "This gives the number of seconds before assuming a calculation has died. If 0 there is no timeout.",
+            },
+        ),
```
litman90, line 896: I don't think this code repetition is needed here.
Look for example the InputFFDebye class.

ipi/inputs/forcefields.py
```diff
@@ -831,3 +833,225 @@ def fetch(self):
...
+        super(InputFFCavPhSocket, self).store(ff)
+
+        self.address.store(ff.socket.address)
+        self.port.store(ff.socket.port)
+        self.timeout.store(ff.socket.timeout)
+        self.slots.store(ff.socket.slots)
+        self.mode.store(ff.socket.mode)
+        self.matching.store(ff.socket.match_mode)
+        self.exit_on_disconnect.store(ff.socket.exit_on_disconnect)
+        self.threaded.store(True)  # hard-coded
```
litman90, line 996: why is this hard-coded here? True is the default value already (line 968)

ipi/inputs/forcefields.py
```diff
@@ -831,3 +833,225 @@ def fetch(self):
...
+            InputAttribute,
+            {
+                "dtype": str,
+                "options": ["auto", "any"],
+                "default": "auto",
+                "help": "Specifies whether requests should be dispatched to any client, or automatically matched to the same client when possible [auto].",
+            },
+        ),
+    }
+
```
litman90, line 959: Some as above regarding code repetition

ipi/engine/forcefields.py
```diff
@@ -1275,3 +1275,490 @@ def poll(self):
...
+        name="",
+        pars=None,
+        dopbc=False,
+        active=np.array([-1]),
+        threaded=True,
+        interface=None,
+        charge_array=None,
+        apply_photon=True,
+        E0=1e-4,
+        omega_c_cminv=3400.0,
```
litman90, line 1494: Please use atomic units here

ipi/engine/forcefields.py
```diff
@@ -1275,3 +1275,490 @@ def poll(self):
...
+    ):
+        """
+        Initialise PhotonDriver
+
+        In this implementation, the photonic masses are set as 1 a.u.
+
+        Args:
+            apply_photon: Determine if applying light-matter interactions
+            E0: varepsilon in the paper, light-matter coupling strength
+            omega_c_cminv: cavity frequency in units of cm-1
```
litman90, line 1297: please use atomic units

ipi/engine/forcefields.py
```diff
@@ -1275,3 +1275,490 @@ def poll(self):
...
+            ph_rep: loose / dense: if the photon coordinates are stored as 2*Nmodes or Nmodes photon "atoms"
```
litman90, line 1298: I am not sure I understand this description, could you clarify it?

ipi/engine/forcefields.py
```diff
@@ -1275,3 +1275,490 @@ def poll(self):
...
+        """
+        self.hartree_to_cminv = 219474.63
```
litman90, line 1300: The unit conversion should not exist here.

ipi/engine/forcefields.py
```diff
@@ -1275,3 +1275,490 @@ def poll(self):
...
+        with self._threadlock:
+            for newreq in newreq_lst:
+                self.requests.append(newreq)
+                self._getallcount += 1
+
+        if not self.threaded:
+            self.poll()
+
+        # sleeps until all the new requests have been evaluated
+        import sys
```
litman90, line 1662: please move this above

examples/lammps/h2o-cavmd/README.md
```diff
@@ -0,0 +1,39 @@
...
+650
+
+       O  2.38972e+00  2.62791e-01  5.65120e+00
+       H  1.41821e+00  1.87797e-01  5.65303e+00
+       H  2.50853e+00  2.13428e-01  4.67340e+00
+       ...
+       L -8.65101e+00  1.11541e+00  4.56823e-01
+       L  5.35376e-01  1.20389e+01 -1.19497e-01
+```
+Here, we have 216 water molecules (648 nuclei) and two photons (labeled as **L**). Please always put the photons at the very end of the xyz file. The mass of each photon is 1.0 atomic units by default. The first photon is coupled to the x direction of the total dipole moment of water, and the second photon is coupled to the y direction. Hence, for the first photon "atom", only the x coordinate is influenced by the molecules; for the second photon "atom", only the y coordinate is influenced by the molecules.
```
litman90, line 39: I guess that third photon is coupled to z, but what happens if we have 4 photons or more
? 
Please add explanation here

ipi/engine/forcefields.py
```diff
@@ -1275,3 +1275,490 @@ def poll(self):
...
+
+        # dimension of independent baths (xy grid points)
+        coeff_x = np.dot(np.transpose(Ekx), self.ftilde_kx)
+        coeff_y = np.dot(np.transpose(Eky), self.ftilde_ky)
+        fx = -np.kron(coeff_x, charge_array_bath)
+        fy = -np.kron(coeff_y, charge_array_bath)
+        return fx, fy
+
+
+class FFCavPhSocket(ForceField):
```
litman90, line 1472: Wouldn't be better to use FFSocket as a parent class to avoid code repetition?

ipi/inputs/forcefields.py
```diff
@@ -831,3 +833,225 @@ def fetch(self):
             active_thresh=self.active_thresh.fetch(),
             active_out=self.active_output.fetch(),
         )
+
+
+class InputFFCavPhSocket(InputForceField):
```
litman90, line 838: wouldn't be better to use InputFFSocket as a parent class to avoid code repetition?
see below

litman90 (changes requested):
Things look really nice.
I left a few minor comments regarding units, and some more relevant ones trying to avoid code repetition. 

I still need to the reg test, I will do it ASAP

TaoELi:
Hi Yair,

Thanks a lot for providing the helpful comments of the code! I have polished the code as you suggested. Please take a look :)

Best,

Tao

ceriottm:
ATM I really don't have the bandwidth to look at the code but I'm glad @litman90  is already on it. Just wanted to thank you a lot for this contribution, that looks really great, and to tell you to also make a PR for the website https://github.com/i-pi/i-pi.github.io to add this to the https://ipi-code.org/about/features/ section

examples/lammps/h2o-cavmd/README.md
```diff
@@ -36,4 +38,4 @@ In short, to perform CavMD simulations, we need to make **two modifications** of
...
+Here, we have 216 water molecules (648 nuclei) and two photons (labeled as **L**). Please always put the photons at the very end of the xyz file. The mass of each photon is 1.0 atomic units by default. We assume a z-direction oriented optical cavity coupled to the molecules, so the k-vector of the cavity photons is along the z-direction, and the cavity photons can polarize along either the x- or y-direction. The first photon is coupled to the x direction of the total dipole moment of water, and the second photon is coupled to the y direction. Hence, for the first photon "atom", only the x coordinate is influenced by the molecules; for the second photon "atom", only the y coordinate is influenced by the molecules. In the current implementation, only two cavity photons are supported (more advanced cavity setups will be included in an updated version of the code).
```
litman90, line 41: Please add a soft exit with this message in case the user use tries to use >2 photons.

ipi/engine/forcefields.py
```diff
@@ -1297,10 +1296,9 @@ def __init__(
             omega_c_cminv: cavity frequency in units of cm-1
```
litman90, line 1296: update line 1297

ipi/engine/forcefields.py
litman90: Just a ping to @taochem, because this point seems to be overlooked

litman90: I would suggest adding ( please check the help string in ipi/inputs/for... for a more verbose explanation )

ipi/engine/forcefields.py
```diff
@@ -1552,6 +1550,12 @@ def calc_dipole_xyz_mm(self, pos, n_bath, charge_array_bath):
         dx_array, dy_array, dz_array = [], [], []
         for idx in range(n_bath):
             pos_bath = pos[ndim_local * idx : ndim_local * (idx + 1)]
+            # check the dimension of charge array
+            if np.size(pos_bath[::3]) != np.size(charge_array_bath):
+                raise ValueError(
```
litman90, line 1555: Please use  softexit here:

Something like:

from ipi.utils.softexit import softexit (add this at the beginning of the file)
softexit.trigger("<Error message>")

litman90 (changes requested):

litman90:
Hi Tao, 
Thanks for addressing the changes. I asked for a few small new tweaks and we should be ready to go (after I create the regtest, which I will do during the weekend).

litman90:
> the website https://github.com/i-pi/i-pi.github.io to add this to the https://ipi-code.org/about/features/ section

I will create an issue to ping other people as well.

TaoELi:
> Hi Tao, Thanks for addressing the changes. I asked for a few small new tweaks and we should be ready to go (after I create the regtest, which I will do during the weekend).

Hi Yair,

Thanks again for reviewing the code again! I have updated the code as you suggested. It is exciting that this CavMD feature is ready to go!

Best,

Tao

litman90 (dismissed):

PR #254, opened 2023-05-07
Bugfix: column number in output of properties with size 1

yotamfe:
Hi,

A small fix to print in the header also the column number of properties that specify size, but that size is 1. An example is the property virial_fq. 

Previously the result was a line "# column --> virial_fq : The scalar product of force and position.", and now it should be something like # column 7 --> virial_fq : The scalar product of force and position.", including the column number. 

This is useful for scripts that parse the output according to the headers, akin to the code in [getproperty.py](https://github.com/i-pi/i-pi/blob/master/tools/py/getproperty.py).

### DIFF
--- a/ipi/engine/outputs.py
+++ b/ipi/engine/outputs.py
@@ -190,15 +190,16 @@ class PropertyOutput(BaseOutput):
             key = getkey(what)
             prop = self.system.properties.property_dict[key]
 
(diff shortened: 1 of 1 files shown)


litman90 (approved):

yotamfe:
Thanks @litman90 for looking into this!

PR #255, opened 2023-05-08
adds amber-compatible netcdf format support for writing trajectories

tjgiese:
Added i-pi support to sander in AmberTools23. The pull request modifies i-pi to write the output trajectories in Amber-compatible NetCDF binary format for analysis with the other Amber-specific tools based on this format. The modifications are largely within ipi/utils/io/backend/io_nc.py, which uses scipy for the backend NetCDF implementation. The format is suitable for writing trajectories for later analysis; however, it does not include (by design) information about elements nor masses (these are typically stored in a separate "parameter file").  Added an example to examples/nc-io/nc-xyz/ which writes a trajectory in xyz and nc formats, and then compares them using i-pi-mux-positions. It's not immediately clear to me how one can write a regression test with the current infrastructure, which appears to use text or numpy-based comparisons.  In principle, one could write a test from a series of commands (involving i-pi and i-pi-mux-positions), as explained in the example.


ceriottm:
Hello. Sorry for the unreasonable lag time. Are you still up to go through a review process to merge this?

tjgiese:
I've been thinking about this more, and I don't think this is a good strategy, so I'm going to close this pull request. Ultimately, I'm going to have to supply users a separate script to transform the amber parameter+coordinate files into a format that ipi can read -- but if I need to supply a script to handle the transformation of the input format, then I might as well have the script also handle the transformation of the output format. In other words, I'll tell people to use the ipi-supported binary format, and then I can externally transform that into netcdf format so they can analyze the trajectories with their standard set of tools. The fundamental difficulty is that the IO formats supported by ipi assume the elements and coordinates are stored in the same file, whereas many simulation packages store them in separate files.  It's easy to ignore the elements when writing the netcdf trajectory, but it's obviously not possible to ignore them when reading the input.

PR #257, opened 2023-05-29
  Diatomic Morse (for exercises)

mahrossi:
Modification to driver so that exercises I am doing become easier.

### DIFF
--- a/drivers/f90/pes/morse.f
+++ b/drivers/f90/pes/morse.f
@@ -7,4 +7,16 @@
          dr=nq-r0
          pot=D*(dexp(-2.*a*dr)-2.*dexp(-a*dr))
          force=-2*a*D*(dexp(-a*dr)-dexp(-2.0*a*dr))*q/nq
-      END SUBROUTINE 
+      END SUBROUTINE
+
+      SUBROUTINE getmorsedia(r0,D,a,q,pot,force)
+         IMPLICIT NONE
+         DOUBLE PRECISION r0, D, a, q(2,3), pot, force(2,3)
+         DOUBLE PRECISION nq, dr, diff(3)
+         diff(:)=(q(1,:)-q(2,:))
+         nq=dsqrt(dot_product(diff,diff))
+         dr=nq-r0
+         pot=D*(dexp(-2.*a*dr)-2.*dexp(-a*dr))
+         force(1,:)=2*a*D*(dexp(-a*dr)-dexp(-2.0*a*dr))*diff/nq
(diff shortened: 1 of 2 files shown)


drivers/f90/pes/morse.f
```diff
@@ -7,4 +7,16 @@ SUBROUTINE getmorse(r0,D,a,q,pot,force)
...
+      SUBROUTINE getmorsedia(r0,D,a,q,pot,force)
+         IMPLICIT NONE
+         DOUBLE PRECISION r0, D, a, q(2,3), pot, force(2,3)
+         DOUBLE PRECISION nq, dr, diff(3)
+         diff(:)=(q(1,:)-q(2,:))
+         nq=dsqrt(dot_product(diff,diff))
+         dr=nq-r0
+         pot=D*(dexp(-2.*a*dr)-2.*dexp(-a*dr))
+         force(1,:)=2*a*D*(dexp(-a*dr)-dexp(-2.0*a*dr))*diff/nq
+         force(2,:)=2*a*D*(dexp(-a*dr)-dexp(-2.0*a*dr))*diff/nq
```
litman90, line 21: Could you double-check? I think there is a minus missing
As reference, the 1D morse is:
      SUBROUTINE getmorse(r0,D,a,q,pot,force)
         IMPLICIT NONE
         DOUBLE PRECISION r0, D, a, q(1,3), pot, force(1,3)
         DOUBLE PRECISION nq, dr
         nq=dsqrt(sum(q*q))
         dr=nq-r0
         pot=D*(dexp(-2.*a*dr)-2.*dexp(-a*dr))
         force=-2*a*D*(dexp(-a*dr)-dexp(-2.0*a*dr))*q/nq
      END SUBROUTINE

litman90 (changes requested):

drivers/f90/pes/morse.f
mahrossi: it is implicit in the "diff" term (that does not exist in the 1D case). I can change for clarity, but before pushing I checked which sign combination would make the simulation stable ;-) the other case explodes, so ... - anyways, we can merge later when I finished all the exercises

mahrossi: EDIT: actually well spotted - the version I pushed was not the version I was running with. All should be fine now. As mentioned, beyond the math all checks pass - dynamics are stable, geometry optimization provides expected geometry and vibrations provide expected frequency.

litman90: cool, thanks for checking

litman90 (approved):

litman90:
@mahrossi I merged it, please delete the branch at your earliest convenience.

PR #258, opened 2023-05-29
Bosonic PIMD in quadratic time

yotamfe:
Hi!

This is an improvement of the bosonic PIMD implementation (Hirshberg et al., PNAS 2019), with an algorithm that scales quadratically rather than cubically. It generates the same potential, forces, and trajectories as before, up to floating point differences, only faster. This is the version of the code that we plan to use for the CECAM tutorial in Tel Aviv. @BarakHirshberg 

We added an automatic test for 3 particles, similar to the existing regression tests.
The documentation of the feature is unchanged.

We tested the code with 4 bosons, 3 bosons, 3 distinguishable particles (but with the same integrator used for bosons), and a mixture of 2 bosons and one distinguishable, all in a harmonic trap, and validated the energy against the analytical result.

Thanks for considering our PR!


yotamfe:
@litman90 Thanks!

litman90:
Hi Yotam, 
if you need this for the school please merge it into this branch https://github.com/i-pi/i-pi/tree/piqm2023
However, keep this PR open, so we can review it and merge it, master, when ready.
Does it make sense for you?

ceriottm:
Hi @yotamfe perhaps since I'll be in Telluride with Barak this is a good time to advance this PR. 
For starters, it needs to be rebased on current master. You can set up your fork so that upstream can always push to it for open PRs, it'd be great if you would set it up that way.

yotamfe:
> Hi @yotamfe perhaps since I'll be in Telluride with Barak this is a good time to advance this PR. For starters, it needs to be rebased on current master. You can set up your fork so that upstream can always push to it for open PRs, it'd be great if you would set it up that way.

Hi @ceriottm,

That would be wonderful! I synced the branch with the upstream's master. I can't find where to change the "Allow edits by maintainers" option - perhaps I don't have the permission and @BarakHirshberg does? I'll try again tomorrow.

Thanks for considering our PR!

ipi/engine/normalmodes.py
```diff
@@ -700,24 +736,27 @@ def get_dynm3(self):
 
     def get_vspring_and_fspring_B(self):
         """
-        Calculates spring forces and potential for bosons.
-        Evaluated using recursion relation from arXiv:1905.090.
+        Calculates spring forces and potential, including exchange effects for the bosons present, if any.
```
ceriottm, line 739: I would restructure this quite heavily. Rather than having a joing v_and_f for bosons, and separate v and f for distinguishable particles, I'd have a single v_and_f for everything, that uses bosons if needed or normal modes otherwise. then you can have a pair of v and f depends that just separate the two components. you can keep the two functions separate if you prefer, but I think at the level of depend objects that would be a better structure.

ipi/engine/normalmodes.py
```diff
@@ -700,24 +736,27 @@ def get_dynm3(self):
...
+                "Bosons must have the same mass, found %s for bosons %s"
+                % (str(masses), str(self.bosons))
+            )
+        boson_mass = masses[0]
+        betaP = 1.0 / (self.nbeads * units.Constants.kb * self.ensemble.temp)
+        # positions of only the boson atoms
+        q = dstrip(self.beads.q).reshape((self.nbeads, self.natoms, 3))[
+            :, self.bosons, :
+        ]
+        self.exchange = ExchangePotential(
```
ceriottm, line 756: I think it'd be better to change the structure of ExchangePotential so you initialize it with the constants (making it perhaps based on depend objects) and then here call it over q.

ipi/engine/normalmodes.py
```diff
@@ -730,15 +769,13 @@ def get_fspring(self):
...
-            f_distinguish = self.transform.nm2b(dstrip(self.fspringnm))
-            # zero force for bosons
-            for boson in self.bosons:
-                f_distinguish[:, 3 * boson : (3 * boson + 3)] = 0.0
+            f_distinguishable = self.transform.nm2b(dstrip(self.fspringnm))
+            f_all = f_distinguishable.reshape((self.nbeads, self.natoms, 3))
+            f_all[:, self.bosons, :] = self.vspring_and_fspring_B[1]
 
-            return f_distinguish + self.vspring_and_fspring_B[1]
+            return f_all.reshape((self.nbeads, 3 * self.natoms))
```
ceriottm, line 778: All this above would be simpler if you computed it in a single f_and_v function.

ipi/engine/properties.py
```diff
@@ -862,6 +862,34 @@ def __init__(self):
...
             },
+            "exchange_distinct_prob": {
+                "dimension": "undefined",
+                "size": 1,
+                "func": self.get_exchange_distinct_prob,
+                "help": "Probability of the distinguishable ring polymer configuration.",
+                "longhelp": """Probability of the distinguishable ring polymer configuration, 
+                               where each atom is has its own separate ring polymer. 
+                               A number between 0 and 1, tends to 1 in high temperatures, which indicates that 
+                               bosonic exchange is negligible""",
```
ceriottm, line 873: Proof-read the above

ipi/engine/properties.py
```diff
@@ -1286,6 +1315,53 @@ def get_kintd(self, atom=""):
             iatom = -1
             latom = atom
 
+        # Should not ask for a property of a subset of atoms of which some are indistinguishables
```
ceriottm, line 1318: of which some -> some of which

ipi/engine/properties.py
```diff
@@ -2613,6 +2686,21 @@ def get_ti_term(self, atom=""):
 
         return ti
 
+    def get_exchange_distinct_prob(self):
+        if not self.nm.exchange:
```
ceriottm, line 2690: I'd throw and error if there are no bosons (and formulate this as `if self.nm.exchange is None` which is more explicit.

ipi/inputs/normalmodes.py
```diff
@@ -75,6 +75,60 @@ def fetch(self):
...
+        {
+            "dtype": str,
+            "default": "index",
+            "options": ["index", "label"],
+            "help": "If 'id' is 'index', then bosonic atoms are specified a list of indices (zero-based). If 'id' is 'label' then specify a list of labels.",
+        },
+    )
+
+    default_label = "BOSONS"
+    default_help = "Deals with the specification of which atoms are have bosonic indistinguishability."
```
ceriottm, line 98: Add more context here, explaining what this does etc.

ceriottm (changes requested):
OK I've a bunch of re-structuring to propose. We can discuss off-github if you want of there's something that doesn't make sense to you.

ceriottm (approved):
This is a big PR and there's always the risk of letting some bugs seep through, but I checked it quite thoroughly and I'm confident that it leaves the bosonic code in better shape than it was before these changes.

PR #263, opened 2023-06-17
Update documentation on tests and examples

ceriottm:


ceriottm:
This is just making sure that the docs don't refer to long-gone nosetests infrastructure. Should be uncontroversial enough not to require review.

PR #266, opened 2023-07-23
PILE thermostats with a different temperature on the centroid

mahrossi:
This commit adds the ability to set a different temperature on the centroid in the PILE thermostats. Would be very gratefull for a review @venkatkapil24 and @ceriottm


ipi/engine/thermostats.py
```diff
@@ -544,7 +561,12 @@ def bind(self, beads=None, atoms=None, pm=None, nm=None, prng=None, fixdof=None)
 
         t = self._thermos[0]
         t.bind(pm=(nm.pnm[0, :], nm.dynm3[0, :]), prng=self.prng, fixdof=fixdof)
-        dpipe(dself.temp, dd(t).temp)
+        if self.pilect > 0.0:  # trying to pipe different temperature here
+            dpipe(dself.pilect, dd(t).temp)
+        else:
+            dpipe(dself.temp, dd(t).temp)
+        #        dpipe(dself.temp, dd(t).temp)
```
ceriottm, line 568: If this work, pls clean up commented out lines like this, so we avoid accumulating noise in the codebase

ipi/inputs/thermostats.py
```diff
@@ -98,6 +99,15 @@ class InputThermoBase(Input):
                 "help": "Scaling for the PILE damping relative to the critical damping. (gamma_k=2*lambda*omega_k",
             },
         ),
+        "pile_centroid_t": (
+            InputValue,
+            {
+                "dtype": float,
+                "default": 0.0,
+                "help": "Option to set a different centroid temperature wrt. that of the ensemble. Only used if value other than 0.0.",
```
ceriottm, line 107: This helpline could be more useful, perhaps mention that this is useful for PIGS

ceriottm (changes requested):
Everything looks reasonable, and - as expected - pretty minimal as an intervention. 
I think it'd be good to clean up comments (and perhaps be a bit more liberal with actual meaningful comments that explain what this pilect does and why. 

I think it'd be nice to also have (1) an example, which we can also use to check this works as intended and (2) properties to print out for the kinetic temperature of different normal modes - these could well be used as diagnostics to ensure there is adiabatic separation.

ceriottm:
Would be good to hear from @venkatkapil24 if this actually works as expected.

venkatkapil24:
I agree with all of @ceriottm's comments. if @mahrossi  could add an example, that would be great as I have an "independent" implementation on my PC.

mahrossi:
Hi, @venkatkapil24 -- if you have an independent implementation, you want to benchmark this against yours? That would be the most helpful thing to do, if you think it is needed.

ipi/inputs/thermostats.py
mahrossi: Regarding the "use in PIGS", I would not add that line @ceriottm -- this is because I have indeed extensively benchmarked it for two models and the realm of "accuracy" is still open. However this should not stop one from adding this functionality to the code because it is indeed a very small change and useful in general (maybe even for different applications).

venkatkapil24:
Tested the implementation. 

Screenshot from 2023-08-15 10-39-58

ipi/inputs/thermostats.py
mahrossi: (but I will make the help string more helpful)

mahrossi:
Superb! Many thanks. I will address the remaining comments in a moment.

ipi/engine/thermostats.py
```diff
@@ -504,11 +519,17 @@ def __init__(self, temp=1.0, dt=1.0, tau=1.0, ethermo=0.0, scale=1.0):
...
            scale: A float used to reduce the intensity of the PILE thermostat if
               required.
+           pilect: centroid mode temperature (if different from ensemble and set by input).
+              Default of 0.0 means it is not used and all modes have equal temperatures.
         """
 
         super(ThermoPILE_G, self).__init__(temp, dt, tau, ethermo)
         dself = dd(self)
         dself.pilescale = depend_value(value=scale, name="pilescale")
+        dself.pilect = depend_value(value=pilect, name="pilect")
```
venkatkapil24, line 529: is this duplicated, considering pile_g inherits from pile_l ?

venkatkapil24 (commented):
happy to merge once we have a regtest.

mahrossi:
Hello, I have a small question that is related to #245 -- if I now try to add an example for this functionality, where would everyone agree it should go? I want to avoid creating a folder under examples/temp/pes/pswater -- would this be a feature? If I look at `features`, they feel like more general methods. Is it ok if I create a folder inside `features` called `thermostats` where we can add `thermostat` examples?

ipi/engine/thermostats.py
mahrossi: I think pilescale is also duplicated. It does not hurt, but it is not nice.

mahrossi:
I overlooked the PIQD folder, which makes more sense. I moved the example now and amended the READMEs. Could this be merged?

ipi/engine/thermostats.py
mahrossi: Done

ceriottm:
> Hello, I have a small question that is related to #245 -- if I now try to add an example for this functionality, where would everyone agree it should go? I want to avoid creating a folder under `examples/temp/pes/pswater` -- would this be a feature? If I look at `features`, they feel like more general methods. Is it ok if I create a folder inside `features` called `thermostats` where we can add `thermostat` examples?

Personally I think the `features/` folder needs some discussion. The inconsistent naming schemes make my eyes bleed, but more substantially there is no clear criterion on how the features should be organized, what level of sub-folders is acceptable, etc.

mahrossi:
> > Hello, I have a small question that is related to #245 -- if I now try to add an example for this functionality, where would everyone agree it should go? I want to avoid creating a folder under `examples/temp/pes/pswater` -- would this be a feature? If I look at `features`, they feel like more general methods. Is it ok if I create a folder inside `features` called `thermostats` where we can add `thermostat` examples?
> 
> Personally I think the `features/` folder needs some discussion. The inconsistent naming schemes make my eyes bleed, but more substantially there is no clear criterion on how the features should be organized, what level of sub-folders is acceptable, etc.

Actually, I have to say I agree. My OCD is also high when looking at the folders now. I'll try to work a bit on that tonight. But that is for #245 and I think this can be merged??

venkatkapil24 (approved):

PR #267, opened 2023-08-17
Fix atom index mapping in atomswap class

ceriottm:
Indices of the selected subsets were not mapped back to the atom indices before the swap. 
This affects the trajectory - but not the converged observables - for simulations that swap all with all, but messes up completely the case in which only a subset is actually considered.


ceriottm:
This is a harmless bugfix, already tested and AFAIK currently only used at COSMO. Will merge quickly so it does not fester

PR #268, opened 2023-08-21
Removed manual.pdf to avoid polluting the website with obsolete info

ceriottm:
The documentation contained a static file with the old pdf manual, which had become _badly_ obsolete.
Given we now have autogenerated reference manual also in the html version, it's better to remove the pdf so we don't give conflicting and misleading info.


mahrossi:
Yes, this is a good idea and completely fine. Please merge.

PR #270, opened 2023-08-22
Examples restructured

litman90:
Only minor changes.
Let's try to keep this updated to master =)


mahrossi:
Why does this have 110 commits? This is looking a bit weird. That branch had been merged.

litman90:
If I look at 'files changed here' everything looks fine.

Regarding the coming count, I think it is because it was squashed before merging. We can kill this branch and open a new one if you think it would be better. Otherwise, we can merge this and continue using this branch for upcoming updates on this front (which we all agree are still needed).

ceriottm:
I think we should probably create a new branch, otherwise we'll get a mess. If you have done any changes, I let you create a new branch from master, move over your changes, and then open a new PR from the new branch. I think moving forward we should delete the branch when we merge, and recreate it if we want to keep the work going.

litman90:
Agree. I will close, delete this branch, and create a fresh PR.

PR #278, opened 2023-10-19
Third round of examples clean-up

ceriottm:
Just a stub for the moment. What I'd like to do:

- [ ]  use lowercase names rather than Camel_Case
- [ ] use underscores rather than hyphen consistently 
- [ ] flatten a bit the folder hierarchy - all the feature_type/feature/lammps_something should be just feature_type/feature
- [ ] use driver whenever possible


ceriottm:
I'll merge this provisionally, before we get divergence in the other feature branches. I'll re-open a new one to continue the clean-up.

PR #281, opened 2023-11-14
Driven Molecular Dynamics

EliaStocco:
Driven Molecular Dynamics
- **author**: Elia Stocco ([elia.stocco@mpsd.mpg.de](mailto:elia.stocco@mpsd.mpg.de))
- **contribution**: the code allows to drive a system using an external time-dependent electric field which couples to the electric dipole of the system. The new `e3nn_pol` driver is also provided, which allows to use a Neural Network to provide the dipole and the Born Effective Charges at each MD step.
- **examples**: two example are provided showing how to drive a vibrational mode of a water molecule. A detailed description of the theory and implementation is provided in the `README.md` file in one of the examples. The examples are in `/examples/features/driven_dynamics/`.
- **limitations**: this implementation can not work with Multiple Time Step Algorithms, and indeed an exception will be raised in case the user tries to use it. The `e3nn_pol` driver is not completely general, i.e. it requires to have some methods defined.


doc/latex/input_ref_idx.tex
mahrossi, line 1: What happened? I think something went wrong in the commit.

drivers/py/pes/e3nn_pol.py
```diff
@@ -0,0 +1,246 @@
...
+        if self.opts["compute-BEC"]:
+            dipole, bec, X = self.model.get_value_and_jac(cell=cell, pos=pos)
+            extras["BEC"] = bec.tolist()
+            # print("sum rule:",bec.sum(dim=0))
+        else:
+            dipole, X = self.model.get(cell=cell, pos=pos)
+
+        extras["dipole"] = dipole.tolist()
+
+        return pot, force, vir, json.dumps(extras)
```
mahrossi, line 246: Ok, we need to discuss this. I propose that we do a new PR soon.

mahrossi (changes requested):

mahrossi:
Hi Elia, many thanks! This is already at a great stage but it is not yet quite ready. My comments will stay here and should be addressed, but I propose that we close this for now, address some issues locally, and then reopen the PR. OK?

PR #283, opened 2023-11-15
Examples cleanup part 4

ceriottm:
Continue the work in #278 in the example folder

- [x]  use lowercase names rather than Camel_Case
- [x]  use underscores rather than hyphen consistently
- [x]  flatten a bit the folder hierarchy - all the feature_type/feature/lammps_something should be just feature_type/feature
- [ ]  use driver whenever possible


litman90:
I think finished points 1. and 2.
Point 3. was overruled by a Slack discussion. @ceriottm @mahrossi Do you remember the final convention we agreed on?
Happy to quickly enforce it. I think it got lost in the Slack history.

ceriottm:
> I think I finished points 1. and 2. Point 3. was overruled by a Slack discussion. @ceriottm @mahrossi Do you remember the final convention we agreed on? Happy to quickly enforce it. I think it got lost in the Slack history.

My recollection is that this convention is what we agreed upon. only have subfolders in the features if there are multiple examples, and only have multiple examples if they add something of substance.  I also think it'd be worth (but perhaps to be left for a fifth PR) to avoid LAMMPS wherever possible. I'd be also be in favor of using a LJ or PH2 example for the constant-stress simulations which is the only thing that for water requires lammps (because of ewald).

litman90:
I flattened the examples folder structure as much as it made sense. This should be ready to merge. @ceriottm, if you agree please merge it. I also moved the profiling folder to ipi_test. 

Modifying all the Lammps examples into driver examples will take a decent amount of effort. I will create an issue so that at some point it is done, but as @ceriottm suggested I would leave it for another PR.

litman90:
(Issue #299)

ceriottm:
rebased on master after boson merge

ceriottm:
I realized I was waiting for the last checkmark but it looks like we all agreed to delay full removal of LAMMPS to a later PR. I nevertheless took out some of the easy ones.

PR #284, opened 2023-11-18
Various profiling-driven optimizations for i-PI 

ceriottm:
Reducing the baseline overhead for running i-PI with a cheapo forcefield. 
This PR focuses on optimizing the simplest NVE MD use case


ceriottm:
This is ready to be checked. There is also a super-duper profiling example but I wouldn't touch the code based on it until this one is merged.

mahrossi (changes requested):
This is deep change in the code. For purposes of clarity, I would like it documented here in this PR (a) what this change does and (b) the rationale behind this change. Say someone would like to add a new depend object now to the code -- how would they do it cleanly and why has it been changed now? 

Further, I would like also some guidance on how to test this. Again, you see, it is passing the regtests, but I am a bit afraid there are corner cases that we are not testing. As this is a big change, I think we need to be careful. What have you tested?

mahrossi:
In fact, amending the online documentation with and explanation about this would be good.

ceriottm:
For the purpose of documenting the changes, the main changes here are:
* switching almost everywhere (softexit being the only exception since it's created only once) from using `threading` to make new threads to `concurrent.futures.ThreadPoolExecutor` that seems to have lower overhead for launching a new worker
* changing the way the "syntactic sugar" around depend objects is implemented. rather than hijacking `__getattribute___` for the parent class (that means a small overhead is added to each member access) I used `@property`-style accessors so the overhead is only for the depend object members. This incidentally also simplifies the usage pattern, because you can access the "real" depend object with `parent._object`.

ceriottm:
For the documentation, if you look at [e6d566e](https://github.com/i-pi/i-pi/pull/284/commits/e6d566e0a9599a395f7e9a407fb95fc0ecf87216) you'll see I already added a usage example and a brief explanation to the docs. Are you thinking of something else/more?

For the errors, I fear it's almost inevitable there will be some. All the regtests pass, and I've made sure there are no more `dobject`, `dself` or `dd` in the codebase. The best we can do is that we all use this branch to run some of our ongoing projects to see if something breaks. Also, we could work on bringing more features (all the suzuki-chin, the fancy estimators, ...) in the `examples/`

mahrossi:
Thanks, also the addition to the docs (older and newer) are great! Tomorrow I will try to run a few things and I hope others can too, just to decrease the chances of bad breaks. Then this should probably be merged and we stay tuned to solve any bugs quickly as soon as they appear.

venkatkapil24:
I am now able to spend some time on this. I found that safe_stride breaks the RESTART file. I will fix it.

venkatkapil24:
I also want to change the latency default from 1e-2 to 1e-4. This has a marginal effect (1 ms for 216 water molecules) on slow drivers (e.g. qtip4pf driver) and makes a big one for fast ones.

mahrossi:
I tested hands-on some rpc, some geops, some instantons, some phonons. All is the same as previously except REALLY MUCH FASTER! From my side, this is ok. Thanks Venkat for spotting the RESTART.

venkatkapil24:
Screenshot 2023-11-27 at 16 54 57

There is a problem here! My i-pi simulations with a large number of atoms (30,000 molecules of water) are failing with the error 


Traceback (most recent call last):
  File "/usr/lib64/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/scratch/snx3000/kvenkat/i-pi-profiling/i-pi-cosmo/ipi/engine/forcefields.py", line 242, in _poll_loop
    self.poll()
  File "/scratch/snx3000/kvenkat/i-pi-profiling/i-pi-cosmo/ipi/engine/forcefields.py", line 354, in poll
    self.socket.poll()
  File "/scratch/snx3000/kvenkat/i-pi-profiling/i-pi-cosmo/ipi/interfaces/sockets.py", line 675, in poll
    self.pool_update()
  File "/scratch/snx3000/kvenkat/i-pi-profiling/i-pi-cosmo/ipi/interfaces/sockets.py", line 736, in pool_update
    driver.get_status()
  File "/scratch/snx3000/kvenkat/i-pi-profiling/i-pi-cosmo/ipi/interfaces/sockets.py", line 302, in get_status
    status = self._getstatus()
  File "/scratch/snx3000/kvenkat/i-pi-profiling/i-pi-cosmo/ipi/interfaces/sockets.py", line 287, in _getstatus
    if not len(reply) == HDRLEN:
UnboundLocalError: local variable 'reply' referenced before assignment

venkatkapil24 (approved):
I am happy with the changes. It looks bug-free.

mahrossi (dismissed):

ceriottm:
Closing because it has been transfered to #294 because of an oh-shit-git situation.

PR #286, opened 2023-11-21
Don't die if the stress is zero

ceriottm:
This is making it harder to run benchmars against master. Changed the error to a warning, should be enough (and ppl should check the pressure...)

### DIFF
--- a/ipi/engine/motion/dynamics.py
+++ b/ipi/engine/motion/dynamics.py
@@ -19,6 +19,7 @@ from ipi.engine.barostats import Barostat
 from ipi.utils.softexit import softexit
 from ipi.utils.messages import warning, verbosity
 
+
 class Dynamics(Motion):
 
     """self (path integral) molecular dynamics class.
@@ -118,8 +119,6 @@ class Dynamics(Motion):
             self.fixatoms = np.zeros(0, int)
         else:
             self.fixatoms = fixatoms
-        
-        self._stresscheck = True
 
     def get_fixdof(self):
         """Calculate the number of fixed degrees of freedom, required for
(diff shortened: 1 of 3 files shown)


mahrossi:
Hi, I don't agree with this change. The reason why this has been added was the measurable amount of computer waste that was generated by i-PI allowing people to run NPT and NST simulations even without stress calculation. Changing it back to this permissive silent warning will make it hard for the users again. Is there another solution?

ceriottm:
> Hi, I don't agree with this change. The reason why this has been added was the measurable amount of computer waste that was generated by i-PI allowing people to run NPT and NST simulations even without stress calculation. Changing it back to this permissive silent warning will make it hard for the users again. Is there another solution?

Not a way that I can think of. It is unfortunate that people run with programs that don't compute the stress, but there are legitimate use cases for running NPT with zero external stress (e.g. to test a barostat). The "right" way to fix the problem would be to communicate (perhaps using the initialize string) that the calculation needs the stress, and then the driver can die with an error if it cannot provide the stress. We could make the warning be printed at each step, which would make it more visible (and simplify the code).

mahrossi:
So, I think that, this is again a case where, in order to make it smoother for developers (i.e., test a barostat), we will make it harder for users. We can control the driver instead, yes, but that means changing the interface right? So we would need to ask all clients to update and IDK about backward compatibility.

ceriottm:
I don't think it's just to make it easier for developers - it's that this is a hacky way to check the simulation is set up well. typical case where this would fail: you use MTS and the fast force computes the stress but you forget to compute it in the ab initio outer step. Stress won't be zero, and everything will be silently broken. 
If the proper fix uses needinit, it might not need to be a breaking change. Only the drivers that request init will receive the information.

PR #287, opened 2023-11-24
Fix for 2-temperature thermostat needed

mahrossi:
Previously, in PR #266 , I had coded the correct temperature rescaling. However, upon new coding I noticed that the code had not been updated. Following the sequence of commits, I can see where it was fixed and then where the fix was silently undone, apparently by the merge conflict that happened in between. Very sorry for this. 

In any case, this is the only way to recover the correct behavior at high temperatures.


ceriottm:
Can we first merge the depend object branch and then apply the change on top of it? That PR touches almost every file, and there will be merge hell if we start merging other stuff before that is properly digested.

mahrossi:
I will merge this because it is a one-line fix and it is a real bugfix. It already has the new depend infra in it.

PR #288, opened 2023-11-24
Opening conversation to enable use of fixatoms and global thermostats on the centroid

mahrossi:
This branch is not ready, but it is as far as I could get now, because an urgent roadblocker appeared.
Importantly I need to fix the thermostat conserved quantity, but also need to make sure this procedure does not break anything else. Putting it here because I think the conversation should be open and this enhances the probability of proper coding in the infrastructure, instead of hacked fixes.


ceriottm:
hello! can you elaborate on what is the problem?

mahrossi:
Currently, there is a stop in the code for the use of fixed atoms and global thermostats on the centroid (in a PIMD setting). This is because of how "ndof" is defined. It was something that happened a while ago, there is a conversation in the i-PI forum where it was discussed (https://groups.google.com/g/ipi-users/c/st4MSR6kIYs/m/txrR8C_ABwAJ). The issue is that the number of fixed degrees of freedom in this case is fixatoms * nbeads, and that does not work for the global thermostat on the centroid. However, the fix has to take into account that if only fixcom is defined, then all is fine.

ceriottm:
OK, I agree we should do something about this. Would it be a good idea to do this in the context of the constraints code? The interaction between those and the thermostats it's something we've been very cavalier about.

mahrossi:
What do you mean by the constraints code? :-) The part that works, or adding more constraints that currently do not work?

PR #290, opened 2023-11-30
Change package name to avoid versioning mess

ceriottm:
__


ceriottm:
This is just to make sure that pip install creates a package with the same name as the library folder.

PR #291, opened 2023-11-30
Added a GH action to release on pypi after release on gh

ceriottm:


ceriottm:
Setting up GH automation for releases so we have more frequently updated pypi version

PR #293, opened 2023-11-30
Updated README to reflect that the pypi package has been renamed

ceriottm:
OK so this is a mini-PR to discuss the renaming of the pypi package. I have also already created a workflow that automatically publishes a new release on pypi and gh when we merge a PR that contains a tag, so after we agree this can be merged let's tag this 2.6.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ python -m pip install git+https://github.com/i-pi/i-pi.git
 
 Last Release::
 ```bash
-pip install -U i-PI
+pip install -U ipi
 ```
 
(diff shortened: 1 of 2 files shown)


ceriottm:
I'll merge this WITHOUT tagging as we have now merged the faster-ipi branch. We should then do a new release once we are ready for a 3.0 beta version, to be sent before submitting the manuscript. I'd appreciate if someone else figured out how to deprecate the `i-PI` release on pypy and redirect to `ipi`

PR #304, opened 2024-01-24
Constrained centroid MD

ceriottm:
This had been implemented since forever, but the option to activate it was hidden. 
This simple PR activates the option and adds a minimal documentation and an example.


mahrossi (commented):
Ok, I understand that `nvt-cc` was implemented already but not documented. Is that the case? Seems like. If so, please merge, looks good.

ceriottm:
Yes, I had implemented it as a side-effect of the planetary model Michael Willatt studied with Stuart, but I tested it and it works fine also on its own.

PR #308, opened 2024-01-30
Fix bug in use of ExchangePotential in NormalModes

yotamfe:
Hi,

This fixes an error that arises when a property such as `exchange_distinct_prob` is used without other properties (such as `conserved`), in which case the NormalModes code acesses the ExchangePotential without first calculating the spring potential, the only place where currently the coordinates are passed to the ExchangePotential. The fix is to initialize with the coordinates also when the potential is created.

Thanks for considering this contribution!


ceriottm:
OK, there are two problems I see here. One superficial - you need to update `black` to the latest pip version to appease the lint CI - one deep: the real problem is that `ExchangePotential` implements the depend machinery in a half-hearted way.

You are making the ExchangePotential object a depend object, but really you should treat it more as a NormalModes object - a foundational class that needs to be initialized and bound, and that will have some properties that can be exposed as depend objects. That way, you can set the dependencies to `q` and all the rest, and you shouldn't have to hack your way around it. Let me know if you see what I'm talking about.

ceriottm:
Moved this development effort to #311

ceriottm:
Moved this development effort to #311

PR #309, opened 2024-02-05
time-dependent NVE integrator

EliaStocco:
# Summary
The code implements the **driving** of the nuclei through an **external time-dependent electric field** (which couples with the dipole of the system) within the `NVE` ensemble.
# Implementation
The equations of motion, which are now time-dependent, are integrated by `EDANVEIntegrator` which inherits from `NVEIntegrator` and adds an extra term to the forces.
# Details
The code can read the dipole of the system if provided by the driver and stored in the `extras` string and JSON formatted.
The derivatives of the dipole w.r.t. the nuclear coordinates, called BECs, can be provided using a txt file (fixed during the simulation) or provided by the driver (in the `extras` string and JSON-formatted).
A specific driver for this task has been implemented in the Pull Request `Adding python driver capable of running example of phonon steering`.
# Limitations
The code works for `nbead>1` too, but not with `Multiple Time Step` algorithms: a warning message is raised if the user tries to use both functionalities.
The code can be easily implemented for `NVT` ensembles as well.


mahrossi (changes requested):

ipi/inputs/eda.py
mahrossi, line 1: This file/class should also be under inputs/motion

ipi/engine/eda.py
```diff
@@ -0,0 +1,350 @@
...
+        self.enstype = enstype
+
+        self._mts_time = depend_value(name="mts_time", value=dstrip(ensemble).time)
+        self._time = ensemble._time
+
+        self.Electric_Field.bind(self, enstype)
+        self.Electric_Dipole.bind(self, ensemble)
+        self.Born_Charges.bind(self, ensemble, enstype)
+
+        # self._Efield = self.Electric_Field#._Efield
```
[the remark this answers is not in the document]
EliaStocco: Done.

ipi/engine/ensembles.py
```diff
@@ -82,6 +82,15 @@ def __init__(
         bweights=None,
         hweights=None,
         time=0.0,
+        # amp=None,
```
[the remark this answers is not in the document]
EliaStocco: Erased.

ipi/engine/motion/dynamics.py
```diff
@@ -105,6 +109,8 @@ def __init__(
             self.integrator = SCIntegrator()
         elif self.enstype == "scnpt":
             self.integrator = SCNPTIntegrator()
+        elif self.enstype == "eda-nve":
+            self.integrator = EDANVEIntegrator()
```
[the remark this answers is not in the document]
EliaStocco: Added one line comment

ipi/engine/properties.py
```diff
@@ -2752,11 +2775,55 @@ def __init__(self):
                 "help": "The momentum trajectories. Will print out one file per bead, unless the bead attribute is set by the user.",
                 "func": (lambda: 1.0 * self.system.beads.p),
             },
+            "becx": {
```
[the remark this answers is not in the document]
EliaStocco: I've added `(x,y,z components in cartesian axes)` to `dipole` and `Efield`.
And `(values go from 0 to 1)` for `Eenvelope`

ipi/inputs/eda.py
```diff
@@ -0,0 +1,156 @@
+"""Creates objects that deal with constant pressure and stress simulations."""
```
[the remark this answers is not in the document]
EliaStocco: Updated to: `Creates objects that deal external time dependent electric field and Born Effective Charges (BEC).`

ipi/inputs/motion/dynamics.py
```diff
@@ -101,6 +113,22 @@ class InputDynamics(InputDictionary):
                 "help": "Number of iterations for each MTS level (including the outer loop, that should in most cases have just one iteration).",
             },
         ),
+        "efield": (
+            InputElectricField,
+            {
+                "default": input_default(factory=ElectricField),
```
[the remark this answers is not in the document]
EliaStocco: Added verbose help string

ipi/inputs/outputs.py
```diff
@@ -141,7 +141,7 @@ class InputTrajectory(InputValue):
             "dtype": str,
             "default": "xyz",
             "help": "The output file format.",
-            "options": ["xyz", "pdb", "ase"],
+            "options": ["xyz", "pdb", "ase", "txt", "text"],
```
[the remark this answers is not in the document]
EliaStocco: I removed these options, I do not need them.

ipi/utils/units.py
```diff
@@ -219,6 +220,28 @@ def mass(cls, label):
...
+        "": 1.00,
+        "automatic": 1.00,
+        "atomic_unit": 1.00,
+        "c/m^2": 57.214766,
+    },
+    "electric-dipole": {  # electron charge/Bohr radius^2
+        "": 1.00,
+        "automatic": 1.00,
+        "atomic_unit": 1.00,
+        # "cm": still to be computed
```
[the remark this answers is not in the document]
EliaStocco: I'va added the conversion for the dipole to eang (electric charge per angstrom). It's the same unit that FHI-aims uses.

ipi/engine/outputs.py
```diff
@@ -345,10 +345,14 @@ def open_stream(self, mode):
             "positions",
             "velocities",
             "forces",
+            "Eforces",
             "extras",
             "extras_component",
             "forces_sc",
             "momenta",
+            "becx",
```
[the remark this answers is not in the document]
EliaStocco: I've added the documentation (comments ahd `help`) for `Eforces` and `becx`, `becy`, and `becz`

ipi/engine/atoms.py
```diff
@@ -85,7 +85,7 @@ class Atoms:
        natoms: The number of atoms.
 
     Depend objects:
-       p: An array giving the components of the atom positions.
+       p: An array giving the components of the atom position.
```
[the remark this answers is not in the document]
EliaStocco: I corrected the description

mahrossi:
@ceriottm - could you review?

examples/features/driven_dynamics/efield-water-beads/.gitignore
ceriottm, line 1: I would avoid local .gitignore files (and actually remove .gitignores that are no strictly necessary)

examples/features/driven_dynamics/efield-water/README.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+#### Dipole and Born Effective Charge tensors
+The dipole $\mathbf{d}$ of the system is not actually computed in this example.
+In the same way we do not *actually* need the energy to integrate Hamilton's equations of motion but only the forces, we just need the Born Effective Charge tensors $Z^*_{ij}$ to run a light-driven simulation, i.e. the derivative of the dipole $\mathbf{d}$ w.r.t. nuclear coordinates $\mathbf{R}$:
+$$
+Z^*_{ij} = \frac{1}{q_e} \frac{\partial d_i}{\partial R_j}
+$$
+
+In this example we will provide the Born Effective Charge tensors $Z^*_{ij}$ using the input file ```bec.txt```, assuming (for semplicity) that their value is constant for each nuclear configuration $\mathbf{R}$.
```
ceriottm, line 59: I think it'd be much better to provide these in the input file rather than through an external file. 
If for whatever reason you want to stick to an external file, explain they can call it as they like and the name is specified in the <bec> field.

examples/features/driven_dynamics/efield-water/README.md
```diff
@@ -0,0 +1,157 @@
...
+
+Let's run the driver (using a UNIX socket) for the energy and forces:
+```bash
+i-pi-driver -u -a driver -m pswater # > pswater.out &
+```
+In case you want to use an INET socket, modify ```input.xml``` as described by the comments therein, and run the driver with:
+```bash
+i-pi-driver -a ipaddress -p 27683 -m pswater # > pswater.out &
+```
+where ```ipaddress``` is your ```ip``` address and the port number (```-p```) should respect the criteria provided by the ```i-PI``` manual.
```
ceriottm, line 151: I don't think these belong here. We don't want to repeat 50 times how to run i-PI, focus on the part that is efield specific

examples/features/driven_dynamics/efield-water-beads/README.md
ceriottm, line 1: It's worth to explain a bit more what's going on here. what is bec.txt, why it's needed, what it means to modify it according to the number of beads, etc.

ipi/engine/__init__.py
```diff
@@ -13,6 +13,7 @@
     "simulation",
     "forces",
     "ensembles",
+    "eda",
```
ceriottm, line 16: could you use something more descriptive?

ipi/engine/motion/dynamics.py
```diff
@@ -118,6 +125,17 @@ def __init__(
         else:
             self.fixatoms = fixatoms
 
+        self.eda_on = False  # whether the dynamics is driven or not
```
ceriottm, line 128: I think there's a lot of unnecessary redundancy in this implementation. It should be possible to implement this as nvtcc - just a change in integrator. Alternatively, if this requires heavier modifications of the MD loop, it'd be best to implement it taking as example the Planetary class. Oveall, I feel it'd be much better to treat this as a different Motion type like planetary than as a further mode of Dynamics, that is already very complex with MTS and all.

ipi/engine/motion/dynamics.py
```diff
@@ -276,6 +311,8 @@ def bind(self, motion):
         self.fixcom = motion.fixcom
         self.fixatoms = motion.fixatoms
         self.enstype = motion.enstype
+        if motion.enstype in EDA.integrators:
+            self.eda = motion.eda
```
ceriottm, line 315: there are way too many of these if statements. everything suggests to me this should be a separate motion class, like planetary or constrained dynamics

ipi/inputs/motion/motion.py
```diff
@@ -107,6 +107,7 @@ class InputMotionBase(Input):
                     "dummy",
                     "scp",
                     "normalmodes",
+                    "BEC",
```
ceriottm, line 110: Why uppercase when it's lowercase everywhere?

ceriottm (changes requested):
My main design issue is that this is much more than a change in integrator, there is the definition of an external field, so it would IMO make more sense as a motion class

mahrossi:
@EliaStocco - check if you can address this soon?

examples/features/driven_dynamics/efield-water/README.md
EliaStocco: I removed this part from the `README.md` file

ipi/engine/__init__.py
EliaStocco: The `EDA` class has not changed name, but is only used "internally", then the name `eda` should not appear anymore in any `__all__` variable or `__init__.py` file.
I created a new `motion` class called `driven_dynamics` for this purpose.

ipi/engine/motion/dynamics.py
EliaStocco: I created a `DrivenDynamics` class, which inherits from `Dynamics`, and can be specified with the keyword `driven_dynamics` in the `input.xml` file (have a look at the examples).
All the rest of my code has not been modified substantially.

ipi/engine/motion/dynamics.py
EliaStocco: These `if` statements no longer exists since the "driven dynamics/EDA" code has been moved into the `DrivenDynamics` class. The only statements of this kind are still in `engine/properties.py`.

ipi/inputs/motion/motion.py
EliaStocco: I just removed `BEC` from this file. It was useless.

examples/features/driven_dynamics/efield-water/README.md
EliaStocco: I would provide the BEC tensors using an external file because they could be really large for bigger systems.
In this `README.md` I have explained how to provide these tensors using an external file and directly into the `input.xml` file.

examples/features/driven_dynamics/efield-water-beads/README.md
EliaStocco: I modified the `README.md` to explain how to provide a `bec.txt` file when `nbeads>1`.
The physical meaning is not clear (to me) right now, and I am not using this functionality yet.
`nbeads>1` will be needed in the future to perform driven dynamics with quantum nuclei.
Right know this example is just a way to check that the code can run with `beads>1`.
For this reason I would not say anything else in this `README.md` file.
Of course, if you do not agree, I can modify the file.

ipi/engine/motion/dynamics.py
EliaStocco: My "old" classes are still in `ipi/engine/motion/eda.py` and `ipi/input/motion/eda.py`.
I created two new files, `ipi/engine/motion/driven_dynamics.py` and `ipi/input/motion/driven_dynamics.py`, to properly implement a new `Motion` class that inherits from `Dynamics`.
I can move all the contents of the `eda.py` files into `driven_dynamics.py`, but if anybody else is interested in implementing something similar, it could be better to have a separate class for `driven dynamics` and then a different file for all possible the ways/flavors to do that (`eda.py` in my case, `some-other-driven-dynamics.py` for somebody else).

examples/features/driven_dynamics/efield-water/README.md
mahrossi: I agree with Elia.

ipi/engine/motion/dynamics.py
mahrossi: I agree that we could see many types of ("time")-driven dynamics.

mahrossi:
hey @ceriottm , we did it - I think this should be merged. can you have a last look?

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+class EDANVEIntegrator(EDAIntegrator, NVEIntegrator):
+    """Integrator object for simulations with constant Number of particles, Volume, and Energy (NVE)
+    using the Electric Dipole Approximation (EDA) when an external electric field is applied.
+    """
+
+    def pstep(self, level):
+        # NVEIntegrator does not use 'super()' within 'pstep'
+        # then we can not use 'super()' here.
+        # We need to call the 'pstep' methods explicitly.
+        NVEIntegrator.pstep(
```
ceriottm, line 173: are you misusing super? should be possible to call the parent function, see the NPTIntegrator
```
        super(NPTIntegrator, self).pstep(level)
```

ipi/engine/motion/dynamics.py
ceriottm: I still see it here though....

ipi/engine/motion/dynamics.py
```diff
@@ -55,6 +56,8 @@ def __init__(
         fixcom=False,
         fixatoms=None,
         nmts=None,
+        efield=None,
+        bec=None,
     ):
```
ceriottm, line 61: is there a reason to have two separate objects for this? seems it'd be appropriate at the very least to condense them in a single eda object

ipi/engine/motion/dynamics.py
```diff
@@ -105,6 +108,9 @@ def __init__(
             self.integrator = SCIntegrator()
         elif self.enstype == "scnpt":
             self.integrator = SCNPTIntegrator()
+        # elif self.enstype == "eda-nve":
+        #     # NVE integrator with an external time-dependent driving (electric field)
+        #     self.integrator = EDANVEIntegrator()
         else:
```
ceriottm, line 114: if this is obsolete, remove, no need to merge comments

ipi/engine/motion/dynamics.py
```diff
@@ -118,6 +124,17 @@ def __init__(
...
+        # if self.enstype in EDA.integrators:
+        #     # if the dynamics is driven, allocate necessary objects
+        #     self.efield = efield
+        #     self.bec = bec
+        #     self.eda = EDA(self.efield, self.bec)
+        #     self.eda_on = True
+        # else:
+        #     # otherwise, keep the class clean and avoid allocating useless things
+        #     pass
+
```
ceriottm, line 137: ditto

ipi/engine/motion/dynamics.py
```diff
@@ -196,6 +213,14 @@ def bind(self, ens, beads, nm, cell, bforce, prng, omaker):
         self.ensemble.add_xlpot(self.barostat._cell_jacobian)
         self.ensemble.add_xlkin(self.barostat._kin)
 
+        # # self.eda has not been allocated if 'self.eda_on' == False
+        # # Attention: call 'self.eda.bind' before 'self.integrator.bind'
+        # if self.eda_on:
+        #     self.eda.bind(self.ensemble, self.enstype)
+
+        # now that the timesteps are decided, we proceed to bind the integrator.
+        self.integrator.bind(self)
```
ceriottm, line 222: this is already called on line 206. clean up

ipi/engine/motion/dynamics.py
```diff
@@ -232,6 +257,15 @@ def step(self, step=None):
...
 
+        # if self.eda_on:
+        #     # Check that these variable are the same.
+        #     # If they are not the same, then there is a bug in the code
+        #     dt = abs(self.ensemble.time - self.integrator.mts_time)
+        #     if dt > 1e-12:
+        #         raise ValueError(
+        #             "The time at which the Electric Field is evaluated is not properly updated!"
+        #         )
+
```
ceriottm, line 268: clean up comments

ipi/engine/motion/eda.py
```diff
@@ -0,0 +1,337 @@
...
+            name="dipole",
+            func=lambda: self._get_dipole(),
+            value=val,
+            dependencies=[ensemble.beads._q],
+        )
+
+        pass
+
+    def store(self, dipole):
+        super().store(dipole)
```
ceriottm, line 152: usually you specift the class and the instance

ipi/engine/motion/eda.py
```diff
@@ -0,0 +1,337 @@
...
+        return np.cos(self.freq * time + self.phase)
+
+
+dproperties(
+    ElectricField,
+    ["amp", "phase", "peak", "sigma", "freq", "Eenvelope", "Efield", "mts_time"],
+)
+
+
+class EDA:
```
ceriottm, line 306: yes I think it'd be better to combine all these objects in one. why passing around three objects that have always to be used together?

ipi/engine/outputs.py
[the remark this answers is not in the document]
ceriottm: given that bec are passed from the driver, can't they just be printed out from extras? what is the need to duplicate here?

ipi/inputs/motion/eda.py
ceriottm, line 1: similar to the case of the EDA object, I think this would be more compact if you didn't have multiple sub-objects doing very little.

ceriottm (changes requested):
Thanks great progress! I think there is still some scope to clean it up and make it clearer and more maintainable. Some choices are a bit puzzling, like the separate time counter. The whole updating manually mts_time and checking if it's equal to time sounds a bit like "fiddling around" without fully grasping what's happening, which worries me slightly.

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+            func=self._eda_forces,
+            value=np.zeros((self.beads.nbeads, self.beads.natoms * 3)),
+            dependencies=dep,
+        )
+        pass
+
+    def pstep(self, level=0):
+        """Velocity Verlet momentum propagator."""
+        self.beads.p += self.EDAforces * self.pdt[level]
+        if dstrip(self.mts_time) == dstrip(self.time):
```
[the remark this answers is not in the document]
mahrossi: Hey @ceriottm , I think you don't know what exactly we are doing, which leads to this and other comments below re. the forces. We can explain but that requires a zoom call, etc. Some of the math is actually in the i-PI paper if you want to check. But we can call.

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+            # the next time this condition will be 'False'
+            # so we will avoid to re-compute the EDAforces
+        pass
+
+    def _eda_forces(self):
+        """Compute the EDA contribution to the forces, i.e. `q_e Z^* @ E(t)`"""
+        Z = dstrip(self.eda.Born_Charges.bec)  # tensor of shape (nbeads,3xNatoms,3)
+        E = dstrip(self.eda.Electric_Field.Efield)  # vector of shape (3)
+        forces = Constants.e * Z @ E  # array of shape (nbeads,3xNatoms)
+        return forces
```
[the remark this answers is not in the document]
mahrossi: Same as above... the hamiltonian is made out of the polarization, which is the object that is position dependent. The BECs are the derivatives of this object wrt. positions, as you would expect in forces...

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+        """Advances the dynamics by one time step"""
+
+        super().step(step)
+        # self.integrator.step(step)
+        # self.ensemble.time += self.dt  # increments internal time
+
+        # Check that these variable are the same.
+        # If they are not the same, then there is a bug in the code
+        dt = abs(self.ensemble.time - self.integrator.mts_time)
+        if dt > 1e-12:
```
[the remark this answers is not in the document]
EliaStocco: They should be always the same at that point of the code.
In case of further changes to the code, I think this a simple way to check if the time-dependent integrators are working properly. 
I can remove or comment these lines, but I think it's a simple check, usefull to future developers.

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+dproperties(DrivenDynamics, ["dt", "nmts", "splitting", "ntemp"])
+
+
+class EDAIntegrator(DummyIntegrator):
+    """Integrator object for simulations using the Electric Dipole Approximation (EDA)
+    when an external electric field is applied.
+    """
+
+    def __init__(self):
+        super().__init__()
```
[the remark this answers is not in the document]
EliaStocco: Right!

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def bind(self, motion):
+        """bind variables"""
+        super().bind(motion)
+
+        self._time = self.eda._time
+        self._mts_time = self.eda._mts_time
```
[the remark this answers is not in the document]
EliaStocco: They are not always identical all over the code! I need two separate variables. But at the end of a complete MD step they should be identical again. This is due to the fact that I within a single MD step the electric field should be evaluated at two different times.
[time-verlet.pdf](https://github.com/i-pi/i-pi/files/14536536/time-verlet.pdf)

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+
+        dep = [
+            self._time,
+            self._mts_time,
+            self.eda.Born_Charges._bec,
+            self.eda.Electric_Field._Efield,
+        ]
```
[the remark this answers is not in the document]
EliaStocco: The Born Charges depend on the nuclear positions, the electric field is just a function of time. I think it makes sense to separate them, especially if there will be in the future more time-dependent motion classes. In this way we have the driving part completely separated from the "dyanmical/position dependent" part.

ipi/engine/motion/dynamics.py
EliaStocco: Right, done

ipi/engine/motion/eda.py
EliaStocco: https://github.com/i-pi/i-pi/pull/309#discussion_r1517507406

ipi/inputs/motion/eda.py
EliaStocco: https://github.com/i-pi/i-pi/pull/309#discussion_r1517507406

ipi/engine/motion/driven_dynamics.py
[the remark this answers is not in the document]
ceriottm: what bugs me is the difference in logic from all the other integrators. the other integrators assume that time is updated after the step, and so act internally as having their own "difference from t=0" tracker. one could modify all integrators to update dynamically time at intermediate steps, or (which IMO would be much easier) make this integrator behave like all the others, knowing that it should evaluate the electric field at t+dt the second time it's called.

ipi/engine/motion/driven_dynamics.py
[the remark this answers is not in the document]
ceriottm: I would agree if the time-dependent integrator was somehow agnostic to the driving mechanism, but as far as I can tell it uses BEC and Efield so if you were to simulate a different driving mechanism you'd have to also introduce another integrator. Am I missing something?

mahrossi:
I think this is not yet perfect but in a much better stage. In any case everything is a work in progress and we will keep pushing enhancements here. @EliaStocco are you happy with the status or did you still wanted to tweak some more given the last discussions? If not, I think we should merge.

EliaStocco:
I think the status of the code is adequate for what you are aiming for.

Of course, improvements are possible (I already have some ideas on how to make them), but they will take time.
Moreover, this new feature is only the first case of many possible different ways of phonon driving (as we discussed with Michele), but the only one currently implemented.

I think that it is necessary to have more examples of phonon driving (which I myself may implement in the near future) so that we have a better idea of how to structure the code.

In the meantime, I will continue to improve the code in a personal repository.

ipi/engine/motion/driven_dynamics.py
ceriottm: Still don't understand the comment. You mean that if you write here 
```
super(EDANVEIntegrator, self).pstep(level) 
``` 
it doesn't work?

ipi/engine/motion/driven_dynamics.py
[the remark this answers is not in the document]
ceriottm: Still think this is unnecessarily complicated, but hey your subroutine.

ipi/engine/motion/driven_dynamics.py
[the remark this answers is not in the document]
ceriottm: I still find this inconsistent with the other implementations but as long as it's self-contained in this file I'm fine with it.

ipi/engine/motion/driven_dynamics.py
```diff
@@ -0,0 +1,177 @@
...
+    """Integrator object for simulations using the Electric Dipole Approximation (EDA)
+    when an external electric field is applied.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def bind(self, motion):
+        """bind variables"""
+        super().bind(motion)
```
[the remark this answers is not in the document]
ceriottm: still unclear what this does

ipi/engine/motion/dynamics.py
ceriottm: ping

ipi/engine/motion/dynamics.py
ceriottm: ping. what are these comments?

ipi/engine/motion/dynamics.py
ceriottm: I think these are leftovers? Please clean up this file from all the comments. You may want to just copy it over from master if you have now moved everything into the driven_dynamics.py

ipi/engine/motion/dynamics.py
ceriottm: ping. what's going on? please bring this back to a clean dynamics.py file.

ipi/engine/outputs.py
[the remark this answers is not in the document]
ceriottm: ping. aren't these part of the stuff that is returned from the driver?

ceriottm (commented):
I don't need to see this again, but please make sure that core files such as dynamics.py are as clean as possible.

mahrossi:
Hi @EliaStocco - I do need to see this again. 

This version here can't work. It has several inconsistencies, including imports that are not valid anymore. So, I cannot trust that this has been benchmarked. I want to see a clean version and a full benchmark of the functionalities with the new code before merging.

EliaStocco:
I cleaned the code from the useless comments and leftovers as pointed out in these comments:
- https://github.com/i-pi/i-pi/pull/309#discussion_r1573017787
- https://github.com/i-pi/i-pi/pull/309#discussion_r1573017871
- https://github.com/i-pi/i-pi/pull/309#discussion_r1573018260
- https://github.com/i-pi/i-pi/pull/309#discussion_r1573020654
- https://github.com/i-pi/i-pi/pull/309#discussion_r1573021424

I removed the EDA class, since it was just a "container" without any active role in the code (https://github.com/i-pi/i-pi/pull/309#discussion_r1573016121).

I am going to modify the integrators design, so that to address these comments:
- https://github.com/i-pi/i-pi/pull/309#discussion_r1573015064
- https://github.com/i-pi/i-pi/pull/309#discussion_r1573016404

I would keep the `becx`, `becy`, `becz` outputs (mentioned here https://github.com/i-pi/i-pi/pull/309#discussion_r1573021882).
They are indeed the outputs returned by the driver, but they formatted in a clean way and automatically saved into a `xyz` file. 
They are not strictly necessary (I can remove them if you want), but I think it's better to have these output nicely formatted instead of read them from a messy file containing `extras` strings.

The idea that I am going to implement is the following:
- `EDANVEIntegrators` will have a `step` method that will call
  - `EDAIntegrator.pstep` : add `Z(q(t))xE(t)` to the momenta
  - `NVEIntegrator.step` : calls the usual `NVE` integrator
  - `EDAIntegrator.pstep`: add `Z(q(t+dt))xE(t+dt)` to the momenta
- the driven dynamics functionality will become compatible with the `MTS` algorithm **but**: 
  -  the time-dependent part can be added **only in the outermost** `MTS` layer:
  - `EDANVEIntegrators.step` will raise an `Exception` if `level != 0` (but level can be higher than one within the `NVE` integrator)

I am also trying to add the time dependent part to an additional `ForceComponent`.

ceriottm:
Thanks for the changes. One thing you also may want to add is a regtest. Over the past weeks I have found these super-useful to spot unintended side-effects so worth to have one, particularly as you're probably going to work on this more in the future.

EliaStocco:
I added a regression test and implemented what I mentioned in the last comments.
What it's still missing (but not crucial at the moment), is to add the time-dependent contribution to the forces to a separate `ForceComponent`.

In the future, I'll work on implementing the following things:
- add the time-dependent contribution to the forces to an additional `ForceComponent`
- implement an `NVT` integrator for driven systems, analogous to the current one for `NVE`
- implement different kinds of electric field profiles (if needed)

mahrossi (changes requested):

PR #311, opened 2024-02-14
Full depend object implementation of ExchangePotential

ceriottm:
Supersedes #308. There is still need to fix the property calculators, but everything seems to be running and to be more in line with the general implementation style of i-PI


ceriottm:
ok following discussion with @yotamfe  I gather this is ready to be merged - I'll do it so we don't have too many open PRs

PR #312, opened 2024-02-15
Add (and update) lost input.xml from an old example

### LINKED ISSUE #264: example/ph2/rpmd lack of input.xml
# this rule creates a run_% subdirectory of rpmd, and executes 16 rpmd runs in it,
# always restarting from the previous run with randomized velocities.
rpmd/run_%/test16.out: 
	cd rpmd; mkdir -p run_$*; cp start-$*.chk run_$*/RESTART; \
	for x in `seq 1 16`; do \
	  sed "s/YYY/"$*"/; s/XXX/$$x/;" input.xml > run_$*/input.xml; \
	  $(IPI) run_$*/input.xml & sleep 25; \
	  $(call run_driver,8,run-$*-$$x); wait; \
	  mv RESTART run_$*/; \
	done  

The Makefile is written in such a way that it appears to require the input.xml file from the rpmd folder using 'sed' command and output it to the run_% folder. However, the input.xml file is not present in the rpmd folder.

Can you provide us with the input.xml file, or should I write the file myself for this part?

ceriottm:
Fixes #264 . This whole folder is probably obsolete, but at least this makes it work.

### DIFF
--- a/examples/temp/ph2/rpmd/input.xml
+++ b/examples/temp/ph2/rpmd/input.xml
@@ -1,0 +1,30 @@
+<simulation verbosity='high'>
+   <output prefix='run_YYY/testXXX'>
+      <trajectory stride='1' filename='vel'> v_centroid </trajectory>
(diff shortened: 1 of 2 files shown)


ceriottm:
Merging this since it's probably dead examples anyway

PR #313, opened 2024-02-19
Adding utilities to parse i-PI output files into more standard python / ase objects

ceriottm:
I keep reimplementing these over and over again, so I thought I'd do it cleanly once and for all. 
I'd like to have one more doing extras parsing but perhaps that can wait until we merge the committee branch.


venkatkapil24:
Also, can you explain these changes to setup.py?

image

ipi/utils/parsing.py
```diff
@@ -0,0 +1,222 @@
...
+                    ) * unit_to_user("force", "ev/ang", 1.0)
+
+                frames.append(frame)
+
+        except EOFError:
+            break
+        except:
+            raise
+
+    return frames
```
mahrossi, line 222: Great feature. I and I am sure many others have a collection of many scripts doing pieces of this. Really glad it is now in a single

setup.py
```diff
@@ -2,19 +2,17 @@
...
+# exclude i-pi-driver that is a compiled fortran file when installing from a local dir
+scripts = [str(p) for p in Path("bin").iterdir() if p.name != "i-pi-driver"]
 
 setup(
     packages=[
         *find_packages(exclude=["ipi_tests*"]),
         "ipi._driver",
         "ipi._driver.pes",
-        "ipi.tests",
-        "ipi.tests.regression_tests",
```
mahrossi, line 12: So, I think people want to do some tests once they installed packages to know whether it is all working in their machine. Why take them all out?

mahrossi (commented):
Really great feature.

setup.py
ceriottm: didn't work unless ran from a gh repo (in which case it picks up tests from the repo folder). never seen a package install tests in the python path.

ceriottm: ping. shall we merge and then possibly open an issue to discuss whether we want to bundle tests, and in that case make sure they actually work?

venkatkapil24 (approved):
If @mahrossi has checked other things, we can merge.

PR #315, opened 2024-02-27
Update the example model for metatensor

Luthaf:
This updates the example model with the latest changes in metatensor. 

It also adds a print to the driver, making the output similar to 

```
$ i-pi-py_driver -a metatensor -u -m metatensor -o nickel.xyz,nickel-lj.pt
This is the Nickel Lennard-Jones model
======================================

A basic implementation of Lennard-Jones using metatensor, to be used for
testing the metatensor integration of various simulation engines.

Model references
----------------

Please cite the following references when using this model:
- about the implementation of this model:
  * https://github.com/lab-cosmo/metatensor
```


Luthaf:
@ceriottm following https://github.com/lab-cosmo/metatensor/pull/539, should I also add the ability to specify dtype/device in the i-PI interface? If yes, what's the best way to do this? Adding a new command line parameter?

ceriottm:
Isn't this specified by the model? I-PI expects double precision and I think there are so many overheads that it wouldn't be worth having the option to run in single precision. So the driver should convert the output to double precision before returning it - I think we're doing that already for PET and MACE. Did I miss something?

ceriottm:
OK, quickly tested, couldn't get it to run. 

```
ImportError: this code is only compatible with metatensor-torch v0.4.x, found version v0.5.0.dev2 at '/home/michele/local/lib/python3.10/site-packages/metatensor/torch/__init__.py'
```

I think the README needs instructions to install the necessary dependencies - if they are there , I couldn't see them.

Luthaf:
> this code is only compatible with metatensor-torch v0.4.x, found version v0.5.0.dev2

You need to install the version of metatensor-torch which is on PyPI, i.e. just `pip install metatensor[torch]`.

ceriottm:
Should I look at this again? Is the pip version of metatensor-torch already including the faster NL?

ceriottm:
@Luthaf  can this be updated to the new NL so I can merge it?

Luthaf:
This should now be updated to use metatensor-torch v0.5, which has the improved NL!

ceriottm (approved):
All good. Still think 0.1s/step for 32 LJ atoms is not the best advertisement for metatensor, but we are going to do some repo madness and it's best to have this merged now.

Luthaf:
> Still think 0.1s/step for 32 LJ atoms is not the best advertisement for metatensor

Well, doing LJ with PyTorch is a bit silly anyway, but it is a simple test for the whole pipeline. We could do something nicer with some proper ML model later.

PR #316, opened 2024-02-29
Removed 'default_hypers_path' parameter from PETCalc call 

HannaTuerk:
Talked to Sergey, the keyword is no longer needed for newest PET version (29.2.2024). I removed the keyword from the pet.py driver for i-pi


ceriottm:
This has been fixed in #334

PR #317, opened 2024-03-05
Another round of performance improvements based on profiling

ceriottm:
Objective here is to target more realistic workflows, such as thermostatted MD and PIMD


ceriottm:
To those worrying about the updated regtests - you are right and this def needs more scrutiny but FWIW I checked that tests broke with the change of the PRNG, which is expected because the "modern" PRNG use a better Gaussian number generator that is not backward compatible. see https://numpy.org/doc/stable/reference/random/new-or-different.html 
I checked that the old tests broke at commit 3f814b142ac9e1f2b61edca81004b921c27c0e01 - the lst before touching the PRNG - and the updated regtests pass at efc74241ac3374723b23c92ca48f4738b3e552a1 - the first commit where the new PRNG is functional

litman90:
> To those worrying about the updated regtests - you are right and this def needs more scrutiny but FWIW I checked that tests broke with the change of the PRNG, which is expected because the "modern" PRNG use a better Gaussian number generator that is not backward compatible. see https://numpy.org/doc/stable/reference/random/new-or-different.html I checked that the old tests broke at commit [3f814b1](https://github.com/i-pi/i-pi/commit/3f814b142ac9e1f2b61edca81004b921c27c0e01) - the lst before touching the PRNG - and the updated regtests pass at [efc7424](https://github.com/i-pi/i-pi/commit/efc74241ac3374723b23c92ca48f4738b3e552a1) - the first commit where the new PRNG is functional

Hi @ceriottm,  can't we update the PNRG in a different and shorter PR? This will allow for a better and faster scrutiny. I am happy to review it carefully. Please tag me when you create it.

ceriottm:
Hello folks, I think this must  reeally be merged *now*, so I can create a 3.0beta tag and we all use it for the "as submitted" benchmarks. If I don't hear from anyone by Tue I'll just merge and hope for the best. As far as I can tell, nothing breaks, besides the regtests I've updated due to the PRNG change.

litman90:
Hi, as discussed in Slack, I will try a few tests to check for physical results in simple/ideal setups. Within the next 7 days.

litman90:
@mahrossi, @venkatkapil24 
You might be the most suited people to have a closer look at the dependence on infrastructure optimization. Do you think you could have a look (at least partially) at it in the next ~ 7 days? If you need more time, how much that would be?

mahrossi:
Hi, I can make time to review this next Tuesday afternoon. I would appreciate if @venkatkapil24 can do it first.

venkatkapil24:
I am also fairly packed (literally too as I am moving places over the weekend). I will try my best to take a look by next Tuesday.

ipi/engine/normalmodes.py
```diff
@@ -87,6 +87,8 @@ def __init__(
         bosons=None,
         dt=1.0,
         nmts=1,
+        fft_threads=1,
+        fft_float32=False,
```
venkatkapil24, line 91: isn't it more efficient to do float32? And if previously things were hardcoded to single precision (float32), as mentioned on slack, then I guess it was tested to provide reasonable accuracy?

ipi/inputs/normalmodes.py
```diff
@@ -159,6 +159,24 @@ class InputNormalModes(Input):
                 "options": ["exact", "cayley", "bab"],
             },
         ),
+        "fft_threads": (
```
venkatkapil24, line 162: Maybe add some guidance on how to set it appropriately on an HPC, considering setting it appropriately can speed up PIMD a fair bit. Is it possible to provide an auto mode which sets it to the number of threads available (OMP_NUM_THREADS)?

ipi/inputs/normalmodes.py
```diff
@@ -159,6 +159,24 @@ class InputNormalModes(Input):
...
+            InputValue,
+            {
+                "dtype": int,
+                "default": 1,
+                "help": "The number of threads to be used for the FFT.",
+                "dimension": None,
+            },
+        ),
+        "fft_float32": (
```
venkatkapil24, line 171: Again, should we use the faster default?

ipi/utils/nmtransform.py
```diff
@@ -364,24 +368,63 @@ def __init__(self, nbeads, natoms, open_paths=None):
...
             # is available
             info(
-                "Import of PyFFTW unsuccessful, using NumPy library instead",
-                verbosity.medium,
+                "Import of PyFFTW unsuccessful, using NumPy library instead. ",
+                verbosity.low,
+            )
+            if self.nbeads > 1:
+                warning(
+                    "Install PyFFTW to avoid slow normal-modes integration",
```
venkatkapil24, line 416: Do we want to make PyFFTW a dependency of i-PI so that people default to using a faster library?

ipi/interfaces/sockets.py
```diff
@@ -299,6 +300,48 @@ def _getstatus(self):
             warning(" @SOCKET:    Unrecognized reply: " + str(reply), verbosity.low)
             return Status.Up
 
+    def _getstatus_direct(self):
+        """Gets driver status. Relies of blocking send/recv, which might lead to timeouts
```
venkatkapil24, line 304: *on

venkatkapil24 (changes requested):

venkatkapil24:
I had a look at everything except the PRNG, which I guess @litman90 agreed to look at.

litman90:
> I had a look at everything except the PRNG, which I guess @litman90 agreed to look at.

I will but please do have a look at that as well.

ipi/engine/barostats.py
```diff
@@ -564,35 +565,31 @@ def pstep(self, level=0):
...
-        dt2 = dt**2
-        dt3 = dt**3 / 3.0
 
         # computes the pressure associated with the forces at each MTS level.
         press = np.trace(self.stress_mts(level)) / 3.0
         self.p += dt * 3.0 * (self.cell.V * press)
 
         # integerates the kinetic part of the pressure with the force at the inner-most level.
         if level == self.nmtslevels - 1:
+            nbeads = self.beads.nbeads
```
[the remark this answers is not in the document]
ceriottm: accessing members of depend objects has a (small but non-entirely-negligible) cost because of the @property decorator. this was even worse in the 2.6.1 depend implementation, but it still makes sense to avoid too many calls to __getitem__ when the same meber is used multiple times

ipi/engine/motion/dynamics.py
```diff
@@ -349,40 +354,26 @@ def pconstraints(self):
...
-            for i in range(3):
-                pcom = p[:, i:na3:3].sum()
-                dens += pcom**2
-                pcom /= Mnb
-                self.beads.p[:, i:na3:3] -= m * pcom
+            vcom = np.sum(p.reshape(-1, 3), axis=0) / Mnb
+            beads.p -= (m3 * vcom).reshape(nb, -1)
 
-            self.ensemble.eens += dens * 0.5 / Mnb
+            self.ensemble.eens += np.sum(vcom**2) * 0.5 * Mnb  # COM kinetic energy
```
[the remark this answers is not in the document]
ceriottm: it's quite distributed, but mainly it's from the array operations. scaling by the masses, summing, subtracting the vcom

ipi/engine/normalmodes.py
ceriottm: no it was not tested I think. this goes back to the very early days. there is a small overhead from doing float64 but it reduces noticeably the drift of the conserved quantity so I think it should be the default

ipi/inputs/normalmodes.py
ceriottm: I tested a bit and it's not trivial, you can "overparallelize" particularly for small number of beads. I think for the moment it's best to leave it as a manual optimization

ipi/inputs/normalmodes.py
ceriottm: as above, I think it was a bad choice to make float32 default, the speedup is not massive and the increase in conserved quantity drift noticeable

ipi/utils/nmtransform.py
ceriottm: IDK, no strong feelings. I like that i-PI has really minimal dependencies

litman90:
I did some ideal gas tests under different NPT/NVT conditions and verified that temperature/volume/pressure are the expected ones according to the equation of state.

I noted, that the conservation of the conserved quantity in NPT is much worse than the one we get in NVT but consistent with the main version of i-pi. (Is this a known fact?)

In summary, this PR is, as far as I can tell, not introducing any bug in NPT/NVT. I suggest that someone with extensive experience running NPT simulations (@mahrossi, @ceriottm) have a quick look at the conserved quantity in NPT runs.
(For example this example: i-pi/ipi_tests/regression_tests/tests/NPT/gas-npzt-PV)

venkatkapil24 (approved):

ceriottm:
OK so to summarize so we can move on and define the simulations to be done to finish the benchmark section, it looks like this does not break anything obvious, and we might have an issue with energy conservation in NPT that is however pre-existing. I'll merge and open an issue about energy conservation in NPT/NST.

PR #319, opened 2024-03-13
A demo of committee models and uncertainty propagation

ceriottm:


ceriottm:
Not perfect, but - after much back&forth with @DavideTisi and @bananenpampe  - good enough to merge. 
We'll update after we have a cleaner metatensor-based potential.

PR #322, opened 2024-04-07
Added an option to override the default prefix for Unix sockets

### LINKED ISSUE #320: UNIX sockets in arbitrary file paths  (instead of `/tmp/ipi_*`)
I'm in a situation where it is possible that a number of different i-pi simulations might be scheduled on the same node at the same time, each with a certain number of clients, and sometimes even within the same SLURM allocation. This creates a problem when using clients which connect via UNIX sockets, since the sockets will be created within the same `/tmp` folder, unless I manually change the addresses in the input file to be unique for each simulation (which is cumbersome). Without too much background knowledge on both the usage of UNIX sockets as well as i-pi, I'm naively tempted to just let each simulation create its own (temporary) directory upon starting i-pi, after which all of the sockets should just get created within that directory. This avoids collisions irrespective of the number of simulations or the number of clients, and does not require changing the input file.

Currently, sockets are forcibly named according to `/tmp/ipi_*`, which makes this approach tricky. What would be the easiest way to proceed? 
https://github.com/i-pi/i-pi/blob/e8601c4c501db7e9f43c5ae30f08446b3a52c7c9/ipi/interfaces/sockets.py#L608

ceriottm:
The default prefix for Unix sockets was hardcoded throughout to "/tmp/ipi_".
This PR added options to both i-pi and the default drivers to override this behavior.

Obviously similar options should be included in the clients, but the default behavior is unchanged so all should be fine. 

Fixes #320 (at least as far as i-PI is concerned, other clients still need to implement their own fix.


mahrossi (dismissed):
The code seems good and harmless to me, there is documentation. It does not conflict with current default work of the code.

mahrossi (approved):

PR #323, opened 2024-04-08
Add elphmod example

janberges:
This example shows how to interface i-PI with [elphmod](https://github.com/janberges/elphmod) as done in [SciPost Phys. **16**, 046 (2024)](https://doi.org/10.21468/SciPostPhys.16.2.046).


litman90:
Hi Mariana, is this ready for review? If I get your confirmation by then I will review it on Monday

mahrossi:
yes! ready to review

examples/clients/elphmod/README.md
```diff
@@ -0,0 +1,42 @@
...
+   Note that the dummy `driver.xyz` required for testing will be overwritten.
+   Git will show the file as modified (or deleted after running `make clean`).
+   Setting up the driver only once in a separate script is more efficient when
+   launching it multiple times later.
+
+3. Now we are ready to run i-PI and the driver, which will communicate via a
+   Unix domain socket, and perform the optimization with `make run`. This will
+   start i-PI in the background, wait until it is ready, and start the driver
+   with `python3 run.py`. i-PI will output the potential energies (`run.out`),
+   positions (`run.pos_0.xyz`), and forces (`run.for_0.xyz`) for all steps.
```
litman90, line 37: same as above

examples/clients/elphmod/run.py
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env python3
+
+import elphmod
+import ipi._driver.driver
+
+driver = elphmod.md.Driver.load("driver.pickle")
+
+ipi._driver.driver.run_driver(unix=True, address="localhost", driver=driver)
```
litman90, line 8: I would leave this as an **alternative** option  adding clearly that it will only work  when i-PI has been installed through pip

examples/clients/elphmod/driver.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+
+import elphmod
+
```
litman90, line 4: Please, explain the main keywords of elphmod used here or add a doc string with the appropriate links to the elphmod documentation

janberges:
Many thanks for considering this example!

I am happy to use `i-pi-driver-py`. As far as I understand, this requires adding a module to `drivers/py/pes/`, which is what kept me from using it so far (to have the whole example in one place). Since `elphmod.md.Driver` has many parameters, it might be the best to have that module load a preinitialized driver from a file whose name is the only argument (see my last commit). Would that be fine with you?

examples/clients/elphmod/README.md
```diff
@@ -0,0 +1,42 @@
...
+This example connects i-PI with [elphmod](https://github.com/janberges/elphmod),
+which provides energy surfaces for distorted systems based on *ab initio* data
+for the symmetric one, using the approximation of a linearized electron-lattice
+coupling. This realizes *model III* by Schobert *et al.*, [SciPost Phys. **16**,
+046 (2024)](https://doi.org/10.21468/SciPostPhys.16.2.046).
+
+elphmod can be installed with `pip` or from `conda-forge`. Since we need the
+Python driver `ipi._driver`, i-PI must be installed with `pip` or `setuptools`;
+running `source env.sh` is not enough. For plotting, `matplotlib` is required.
+
```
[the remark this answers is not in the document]
janberges: I agree. I use `i-pi-driver-py` now, see my comment below. (As a consequence, the example will only work with the next PyPI release of i-PI of course.)

examples/clients/elphmod/run.py
janberges: For now, I have fully removed this option to keep it simple. If you think it is useful, I can bring back the removed script `run.py` and add an alternative phony target to the makefile that uses it.

examples/clients/elphmod/driver.py
janberges: Thanks for the hint! Is the minimal docstring with the link to the documentation of the relevant module enough? (I have chosen the Markdown format because it is used in some other docstrings of the repository.)

mahrossi: I would prefer to leave as Jan has it now, because describing all elphmod keywords here would be a repetition of work (that has been nicely done elsewhere).

mahrossi:
Hey @janberges thanks for addressing all points, real great. So, one thing with the current pre-loading is that it becomes less transparent (at least to me), what would a user need to do if they wanted to use elphmod with their own (new) setup. I think this is a good compromise overall, but honestly, it would also help to actually have a more transparent setup. Before, it was more transparent. Is there a way to at least describe in the README what a user would have to do instead?

examples/clients/elphmod/driver.py
litman90: yes, that is enough thanks

litman90 (commented):
Please import 'sys' in drivers/py/pes/elphmod.py 
After that, we will be ready to merge. Thanks

examples/clients/elphmod/README.md
janberges: The numbered instructions are now only for the `i-pi-driver-py` way of doing this. Anything else is discussed at the end.

examples/clients/elphmod/run.py
janberges: OK, it is probably better to leave this as an alternative option, because it shows that also `elphmod.md.Driver` on its own (without the wrapper in `drivers/py/pes/elphmod.py`) works like the drivers defined in `drivers/py/pes/`. It can be tried using the new phony target `run2`, which is not used by default.

janberges:
Thanks! (It seems I only ran `make pretty`, sorry.) To hopefully make the reasons behind creating an intermediate driver file more transparent, I have added a comment at the end of the README. Is it now clearer that one can define a custom setup as in the script `driver.py`?

litman90 (approved):

litman90:
Thanks Jan for your contribution.
From my side, it is ready to merge (provided all the tests pass).
@mahrossi, I am not sure if you wanted to ask Jan something more before merging.
Feel free to merge if you are satisfied.

mahrossi:
I did ask in my comment above.

PR #328, opened 2024-04-16
Fix nasty queueing bug

ceriottm:
@DavideTisi , try this branch, it should fix the PIMD problem

### DIFF
--- a/ipi/engine/forces.py
+++ b/ipi/engine/forces.py
@@ -428,7 +428,7 @@ class ForceComponent:
 
         is_tainted = False
         for b in range(self.nbeads):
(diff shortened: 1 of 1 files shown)


mahrossi:
the order of the or condition matters?!?!?

DavideTisi:
yes this is python magic ahaha
If the first condition is True, thus the whole `or` is `True` it does not evaluate the second

ceriottm:
@DavideTisi  if you confirm that this works I would merge

DavideTisi:
Yes yes it works, it scales as it should

PR #330, opened 2024-04-20
Better PLUMED interface

ceriottm:
Adds functionality to retrieve CVs from PLUMED. 

Also updating examples and adding two regtests, that have to be run manually given there's no PLUMED on the CI.


ceriottm:
@venkatkapil24  I need this to clean up the committee example - if you can't look at it quickly let me know and I'll just merge it, should be harmless outside of the PLUMED niche.

examples/features/metadynamics/pimd_metadynamics_zundel/input.xml
```diff
@@ -33,7 +35,9 @@
       <ensemble>
          <temperature units="kelvin"> 300.0 </temperature>
          <bias>
-            <force forcefield="plumed" nbeads="1"></force>
+           <force forcefield="plumed" nbeads="1">
+             <force_extras> [ doo, mtd.bias ] </force_extras>
```
venkatkapil24, line 39: So this uses ring polymer "expansion" allowing beadwise printout of extras? What if I am running PIMD and I want to print out the extras for the centroid?

examples/features/metadynamics/README.md
```diff
@@ -15,3 +15,5 @@ i-pi--driver -u -h zundel -m zundel
 ```
 
 However, they require having PLUMED libraries and Python bindings in the appropriate system paths.
+
+Note also in `pimd_metadynamics` how to retrieve CVs computed PLUMED-side back into i-PI.
```
venkatkapil24, line 19: Can you be a bit more verbose here that the CVs or CV attributes can be specified in <plumedextras> </plumedextras> and printed out as a trajectory by defining the `extras_type` attribute of `extras bias`.

venkatkapil24: Can you also explain to me if we need to specify the variables in forces_extras that need to be printed?

venkatkapil24: So, if I understand correctly,  if one does stuff classically, then the variables need not be defined in forces_extras. However, for PI simulations, the user will need to as trajectories are only printed for the beads?

examples/features/metadynamics/pimd_metadynamics_zundel/input.xml
ceriottm: Well, you have to compute it, so you would add a "silent" force component
```
<force forcefield="plumed" nbeads="1" weight="0.0"></force>
```
and reference it from the `<outputs>` section with `potential_component(idx)` properties and trajs. That way it'll be computed only when it's required to print it out.

examples/features/metadynamics/README.md
ceriottm: Yes. `forces_extras` is needed to tell i-PI that the named extras (1) have a numerical value and (2) can be Fourier interpolated/extrapolated to compute them for the full ring polymer. The docs for forces_extras says it well
```
A list of quantities that should be extracted from the ‘extra’ string returned by the forcefield, that are then treated the same way as energy or forces – that is treated as a numerical, physical quantity, interpolated when changing the number of PI replicas. Same quantities from different force components are summed as well. The names should correspond to entries in the JSON-formatted extra string.
```
Maybe we can use a more telling name as `extras_interpolation` ?

venkatkapil24: yeah, or how about 'ring_polymer_interpolation_extras' or 'ring_polymer_extras'?

examples/features/metadynamics/pimd_metadynamics_zundel/input.xml
venkatkapil24: I guess this is not entirely clear to me (an experienced user), so I think it will also be unclear for other practitioners. My confusion is twofold:

1. How one can access the extras from a given 'force component'. Do we have an example for this? 

2. Isn't the extras like any trajectory-compatible quantity represented internally beadwise? So even if `<force forcefield="plumed" nbeads="1" weight="0.0"></force>` calculates the forcefield (forces and extras), wouldn't the print option only be available for beads? I don't see, other than the case for classical MD where beads and centroid are the same, how one can print out extras for a centroid.

Maybe this use case needs an example.

examples/features/metadynamics/README.md
ceriottm: ok, I have now pushed this and I also ended up adding `forces_component` output. In analogy with `pot_component`, by default it'll contract to the centroid so this should cover your use case. PLS test that it works before we merge - in all events I've given up on having a finalized ensemble-demo before submission.

venkatkapil24:
Final question: if the quantities specified in plumed extras [doo, co1.lessthan, co2.lessthan, mtd.bias ] are not specified in forces_interpolate, will the centroid quantities still be accessible in the raw json format?

venkatkapil24:
I have tested the force component print out. I do not have access to plumed right now so I am assuming @ceriottm has checked that.

venkatkapil24:
I have tested the force component printout. I do not have access to Plumed right now, so I am assuming @ceriottm has checked that.

examples/features/metadynamics/pimd_metadynamics_zundel/input.xml
ceriottm: yes, there are examples in `examples/temp/pes/pswater/combine-extras`. still need to be ported somewhere inside the "canonical" examples - feel free to help ^_^

PR #334, opened 2024-05-14
Update PET driver to work with current code

sirmarcel:
This makes the PET driver compatible with the current iteration of the PET code. It also adds some small amount of future-proofing by allowing to pass through arbitrary keyword arguments.

### DIFF
--- a/drivers/py/pes/pet.py
+++ b/drivers/py/pes/pet.py
@@ -26,11 +26,13 @@ __DRIVER_CLASS__ = "PET_driver"
 class PET_driver(Dummy_driver):
     def __init__(self, args=None, verbose=False):
         self.error_msg = """
-The PET driver requires specification of a .json model file fitted with 
-    the PET tools, and a template file that describes the chemical makeup of 
-the structure. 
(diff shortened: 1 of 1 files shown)


drivers/py/pes/pet.py
```diff
@@ -26,11 +26,13 @@
...
-    the PET tools, and a template file that describes the chemical makeup of 
-the structure. 
+The PET driver requires (a) a path to the results/experiment_name folder emitted by pet_train
+                        (b) a path to an ase.io.read-able file with a prototype structure
 
-Example: python driver.py -m pet -u -o model.json,template.xyz
+Other arguments to the pet.SingleStructCalculator class can be optionally
+supplied in key=value form after the required arguments.
+
+Example: python driver.py -m pet -u -o "path/to/results/name,template.xyz,device=cuda"
```
ceriottm, line 35: @sirmarcel love this, maybe at a later stage we can move it into the base driver class, as I guess any PES can benefit from this flexibility

ceriottm (approved):

PR #338, opened 2024-05-31
Ensure that sockets_prefix is passed to run_driver

sirmarcel:
This makes sure that the command line argument `-S` actually works as intended when passed to the python driver.


mahrossi (approved):

mahrossi:
yep, was missing - I will merge.

PR #340, opened 2024-06-04
Fix aspects of interface between i-Pi and PLUMED that are not working

gtribello:
There are two problems with the interface between i-Pi and PLUMED at the moment:

(1) The timestep is set incorrectly.  It is actually not possible to set the timestep correctly.  I have thus deleted the command to set the PLUMED timestep here. 

(2) The cell appears to be passed incorrect.  If you pass a deep copy of the NumPy array containing the cell as I do here I think the problem is fixed.

If you output a COLVAR file with PLUMED with this commit the output will look like this:

```
#! FIELDS time CV
0.00000  1.0
0.00000  2.0
0.00000  3.0
````

In other words, the CV column will be reported correctly but all the times will be zero as the timestep is not passed from the MD code.

### DIFF
--- a/ipi/engine/forcefields.py
+++ b/ipi/engine/forcefields.py
@@ -705,13 +705,15 @@ class FFPlumed(FFEval):
         self.plumed.cmd("setMDEngine", "i-pi")
         self.plumed.cmd("setPlumedDat", self.plumeddat)
         self.plumed.cmd("setNatoms", self.natoms)
+        timeunit = 2.4188843e-05
+        self.plumed.cmd("setTimestep", 1/timeunit)
         self.plumed.cmd(
             "setMDEnergyUnits", 2625.4996
         )  # Pass a pointer to the conversion factor between the energy unit used in your code and kJ mol-1
         self.plumed.cmd(
             "setMDLengthUnits", 0.052917721
         )  # Pass a pointer to the conversion factor between the length unit used in your code and nm
-        self.plumed.cmd("setMDTimeUnits", 2.4188843e-05)
+        self.plumed.cmd("setMDTimeUnits", timeunit)
         self.plumedrestart = False
         if self.plumedstep > 0:
             # we are restarting, signal that PLUMED should continue
--- a/ipi/engine/forcefields.py
+++ b/ipi/engine/forcefields.py
@@ -705,7 +705,11 @@ class FFPlumed(FFEval):
         self.plumed.cmd("setMDEngine", "i-pi")
         self.plumed.cmd("setPlumedDat", self.plumeddat)
         self.plumed.cmd("setNatoms", self.natoms)
-        timeunit = 2.4188843e-05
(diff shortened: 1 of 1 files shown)


ceriottm:
Can we set the timestep to 1.0? I think it'd be better to have a "call counter" than have no information at all.

gtribello:
To get that to work you would need to set the timestep to the inverse of the time unit.  Then when it is converted you should get one.

mahrossi:
Hi, I agree that a counter would be good - or Gareth's suggestion with the inverse units, but done in an automatic fashion such that the user does not need to care about it.

In addition, can we have a plumed regression test now that this is fixed? It would be nice to include that in this PR.

ceriottm:
Problem with a regtest is that we'd need PLUMED in the CI

gtribello:
Hello folks

I have updated this PR so that the timestep is always one.  You thus get a COLVAR file that looks like this:

````
#! FIELDS time CV
1.00000  1.0
2.00000  2.0
3.00000  3.0
````

On testing, the reason I am looking into this is to try to build this site here:

https://plumed-testcenter.github.io

For testing the interfaces between plumed and the various MD codes that call it.

mahrossi:
> Problem with a regtest is that we'd need PLUMED in the CI

Yes, I know - I was thinking we could just pip install it in the CI -- but I read the PLUMED page now and it seems to be a bit more convoluted than that? https://pypi.org/project/plumed/ ?

mahrossi:
@gtribello - could you please "black" the files you changed?

gtribello:
OK black has been run.

gtribello:
Hello folks

On a separate note, I am not sure if NPT simulation with I-Pi and PLUMED are set up correctly.  The tests that we run here:

https://plumed-testcenter.github.io/tests/i-pi/virial_v2.9.html

are not passing.  The details of the test that we do are described on that page.  I don't really know what to do to investigate this further, sorry.  I tried printing things from I-Pi and plumed calculations but I didn't really understand what I was looking at sorry.  It would be nice to know if this is a real problem or not.  Thanks.

ceriottm:
> Hello folks
> 
> On a separate note, I am not sure if NPT simulation with I-Pi and PLUMED are set up correctly. The tests that we run here:
> 
> https://plumed-testcenter.github.io/tests/i-pi/virial_v2.9.html
> 
> are not passing. The details of the test that we do are described on that page. I don't really know what to do to investigate this further, sorry. I tried printing things from I-Pi and plumed calculations but I didn't really understand what I was looking at sorry. It would be nice to know if this is a real problem or not. Thanks.

Let's start to merge this, and open an issue re the barostat to investigate this further.

gtribello:
OK issue has been created.

ceriottm (approved):
Ooops. Thought I had merged already. Let's see if it still works out.

PR #341, opened 2024-06-04
Add rudimentary parsing of literals to PET driver

sirmarcel:
This is needed to pass non-strings such as `None`, `True`, or `False` into the `SingleStructCalculator`.

I've drawn the line at floats and ints, because it is not so easy to programmatically distinguish them, and it's not always clear what the intended type is supposed to be.


ceriottm:
Good enough for the moment, we can do this more systematically in a future PR. Would you open an issue about streamlining option parsing for the python driver?

sirmarcel:
Thanks, opened the issue #342 👍

PR #344, opened 2024-06-06
apply string replace from longest to shortest label

### LINKED ISSUE #343: i-PI system template matching substitutes substrings
Consider the following input file:

```
<simulation verbosity="high">
  <ffsocket mode="unix" name="PlumedHamiltonian1" pbc="False">
    <timeout>10.0</timeout>
    <exit_on_disconnect> TRUE </exit_on_disconnect>
    <address>plumedhamiltonian1</address>
  </ffsocket>
  <ffsocket mode="unix" name="PlumedHamiltonian10" pbc="False">
    <timeout>10.0</timeout>
    <exit_on_disconnect> TRUE </exit_on_disconnect>
    <address>plumedhamiltonian10</address>
  </ffsocket>
  <output prefix="output">
    <checkpoint filename="checkpoint" stride="250" overwrite="True" />
  </output>
  <system_template>
    <labels>[ INDEX, TEMP, PLUMEDHAMILTONIAN1, PLUMEDHAMILTONIAN10 ]</labels>
    <instance>[ 98, 1000.0, 1.0, 0.0, ]</instance>
    <instance>[ 98, 1000.0, 0.0, 1.0, ]</instance>
    <template>
      <system prefix="walker-INDEX">
        <initialize nbeads="1">
          <file mode="ase" cell_units="angstrom"> start_INDEX.xyz </file>
          <velocities mode="thermal" units="kelvin"> TEMP </velocities>
        </initialize>
        <motion mode="dynamics">
          <dynamics mode="nvt">
            <timestep units="femtosecond">0.5</timestep>
            <thermostat mode="langevin">
              <tau units="femtosecond">100</tau>
            </thermostat>
          </dynamics>
          <fixcom> True </fixcom>
        </motion>
        <ensemble>
          <temperature units="kelvin">TEMP</temperature>
        </ensemble>
        <forces>
          <force forcefield="PlumedHamiltonian1" weight="PLUMEDHAMILTONIAN1" />
          <force forcefield="PlumedHamiltonian10" weight="PLUMEDHAMILTONIAN10" />
        </forces>
      </system>
    </template>
  </system_template>
  <smotion mode="remd">
    <remd>
      <stride>1000</stride>
      <krescale>True</krescale>
      <swapfile>replica_exchange</swapfile>
    </remd>
  </smotion>
  <total_steps> 5000 </total_steps>
  <prng>
    <seed> 12345 </seed>
  </prng>
</simulation>
```
The intended substitution would be that `PLUMEDHAMILTONIAN1`  would get matched only to `weight="PLUMEDHAMILTONIAN1"`, but in reality, due to [this line](https://github.com/i-pi/i-pi/blob/453fa33de9867aa2d18f62f9039ea148ca038249/ipi/inputs/system.py#L92), he would also blindly substitute it into the second force since it just replaces any occurrence. Woulnd't it make more sense to first find all occurrences of the template parameter, but filter out those for which the enclosing double quotes include other characters, such that when only a substring is matched, it actually ignores this? 
Alternatively, as a user, I need to guarantee there is absolutely no chance of substring matching. While that's possible, it seems like something which would be safer to fix on the i-PI side ...

svandenhaute:
fixes #343 according to @ceriottm 's suggestion. Probably a safer solution vs using regexp.

### DIFF
--- a/ipi/inputs/system.py
+++ b/ipi/inputs/system.py
@@ -4,8 +4,6 @@
 # i-PI Copyright (C) 2014-2015 i-PI developers
 # See the "licenses" directory for full license information.
 
-import re
-
 import ipi.engine.system
 from ipi.utils.depend import *
 from ipi.utils.inputvalue import *
--- a/ipi/inputs/system.py
+++ b/ipi/inputs/system.py
@@ -4,6 +4,7 @@
 # i-PI Copyright (C) 2014-2015 i-PI developers
 # See the "licenses" directory for full license information.
 
+import re
 
 import ipi.engine.system
 from ipi.utils.depend import *
@@ -88,8 +89,9 @@ class InputSysTemplate(Input):
                 sys = template
                 if len(labels) != len(ins):
                     raise ValueError("Labels and instance length mismatch")
-                for l in range(len(ins)):  # string replacement within the template
-                    sys = sys.replace(labels[l], ins[l])
+                for label, to_insert in sorted(zip(labels, ins), reverse=True):
+                    # sort from longest to smallest label, to avoid replacing 'string' in 'string1'
+                    sys = sys.replace(label, to_insert)
                 print("Generating system from template: \n", sys)
                 xsys = xml_parse_string(sys)  # parses the string to an XML object
                 isys = InputSystem()


venkatkapil24:
@svandenhaute can you use i-pi-style in the i-pi parent directory to ensure lint is happy?

venkatkapil24 (approved):

PR #347, opened 2024-06-09
add LICENSE.md file before v3.0 release

litman90:
Add a LICENSE.md file explaining that the code within the driver folder is published within the MIT license and the rest under the GPL license.
@ceriottm, if you want to change the licence, now would be a good time. Otherwise please merge this.


ceriottm (dismissed):

ceriottm (commented):
OK Sorry wait a sec, why do we need this in addition to the existing license files?

litman90:
Because, as far as I can tell, it is not specified clearly what a GPL licence, is and what's MIT licence.

litman90:
and I thought this was a good place to make it explicit.
Should I merge, or do you prefer doing it in differently?

ceriottm:
OK so the idea, that was discussed and agreed upon ages ago, is that the whole project is DUAL licensed - so one can pick whatever license suits them best.

mahrossi:
Sorry I missed this - yes, I confirm we agreed that all of it should be dual-licensed. I heard criticisms about this from external sources. But I am not sure we can change that now.

litman90:
As the owners of the code, if we all agree, we can change it ( I mean legally speaking) .
We can discuss it quickly tomorrow.

litman90:
Adding LICENSE.md  file as discussed in the last coordinator's meeting.

PR #351, opened 2024-06-16
Gh page update2

litman90:
This OR updates mainly the demos/para-h2-tutorial/
It cleans the explanations, adds many missing pieces, fixes the inputs, etc.
Again, there is room for improvement, but at least it works now.
@yotamfe will have a look so a follow-up PR with more corrections/modifications might appear soon.


litman90:
I proceed to merge this since it contains only non-critical doc-related fixes.

PR #353, opened 2024-06-18
Fixed buggy env.sh

ceriottm:
env.sh was broken, we didn't notice because we all have a proper pip install to fall back to. 
This should work on bash, dash and zsh


litman90:
I confirm it works in Ubuntu 22.04.4 LTS

DavideTisi (approved):
I works on BASH

mahrossi:
works in zsh

PR #355, opened 2024-06-19
  ffsocket pbc option is now false as default!

mahrossi:
This changes the default of the "pbc" input parameter in ffsocket to *False*.

  Note that this only controls whether positions are shifted and wrapped before
  sending to client or not. Simulation continues to be periodic regardless.


litman90:
I think it is important that you summarise in this PR  the rationale for this change of default and state clearly the codes you are aware should not use the default.
(This is just to not rely on our faulty memories, nor on our ability to find them in the slack channel).

mahrossi:
At some point in the past, it was important to have pbc=True for LAMMPS clients. There were problems if molecules were wandering more than 2 lattice constants away from the original box (@ceriottm could be more quantitative here, as the memory of the problem escapes me). For all other client codes, only one of two options were possible: (1) they did not care whether pbc was true or false and worked fine regardless or (2) they would break down either completely or in corner cases if pbc was true. Elphmod, for example, breaks completely if it is true. FHI-aims breaks in corner cases - when there was an atom right at the edge of the box. Then the application of pbc by i-PI and the subsequent application of pbc by FHI-aims made atoms go step by step one more box away until their coordinates made no sense. I know these because I use them frequently, but I am sure other clients could have similar corner cases that went unreported. This led to a situation that we were most of the time "for safety" using pbc=False anyways. Recently, some new implementation in the group in the driver code _again_ had serious issues because of the silent "pbc=true" flag. To finish it off, @ceriottm revamped the LAMMPS interface last month and he reports that the new implementation should have solved the old issue. He can be more quantitative here again. So we set the default to false and I hope @litman90 is satisfied with the long text! ;-)

ceriottm:
someone needs to please (1) check quickly that (1) new LAMMPS gives sensible results if you pass an atom that is like 10 boxes away from the origin; (2) add a couple of lines to the documentation and/or make the docstring for the pbc option more verbose. otherwise all looks good

mahrossi:
hey Michele, but even if LAMMPS does not give a sensible result, one can always set pbc=True for that case and the old behavior is unchanged. I am only changing the default.

yotamfe:
Hi, sorry for barging in - if there is a "silent" change of behavior w.r.t. the previous version, would it make sense to provide an "upgrade" script that users are supposed to run to convert their inputs from the old version to the new one? Or alternatively to make this parameter required (in which case users might write their own upgrade scripts in case they relied on the default behavior)?

mahrossi:
I don't think an "upgrade script" is possible or makes sense in this case. Note that in principle all simulations will not notice this change in behaviour - either because they were already using pbc=false explicitly because they needed it, or because they don't suffer with it at all. No regtest was really any different with this, for example (except for a discrepancy at 1e-14 level on numerics of one example which I am not sure comes from this change). 

Making the flag mandatory would break a lot of simulations and require that everyone tests. I can do that, if people agree. I would also be happy if you @yotamfe would contribute with such an upgrade script if you have a good idea for it.

yotamfe:
I've added an example of I'd had in mind, for the sake of users who might be relying on the default value to be pbc=True. If the same input file can mean different things in two versions (because the default value changed), then I think it also makes sense to also add an (optional) version field to the input, the script can add it too if we want. Perhaps having a version field is a good idea anyways, in anticipation of similar changes in the future.

mahrossi:
Thanks Yotam. I will add the version field to the Hackathon TODO list. I am still unsure about the script as we discussed in parallel. Let's wait for the LAMMPS test and then try to merge.

mahrossi:
@DavideTisi has confirmed that LAMMPS now works seamlessly with the pbc=false default, thus reducing to zero the cases we know of that could have an issue with the new default. of course we do not know of absolutely all possible cases that individual users might have! :-)

ceriottm:
One thing - I think that having a convert script without a input versioning and a set of schemas is more confusion than help so maybe we can wait until the hackathon to do a proper job

mahrossi:
Ok. It seems to me that the script does not hurt, but I removed all reference to "upgrade" - to try to avoid people thinking they need to do this to all input files - and I also added a long explanation to the script. Hoping people read it. Then, I think it is ok if it survives where it is, and we can repurpose it when we implement versioning of the inputs.

ceriottm (approved):

PR #356, opened 2024-06-20
Carry on and display more meaningful errors for py driver

ceriottm:
When importing one of the PES files fails the python driver currently dies with a cryptic message. This small PR lets the driver continue, and prints more meaningful error messages that can help debug the faulty PES.

### DIFF
--- a/drivers/py/pes/__init__.py
+++ b/drivers/py/pes/__init__.py
@@ -14,7 +14,7 @@ for loader, module_name, is_pkg in pkgutil.iter_modules(__path__):
     # Import the module
     try:
         module = importlib.import_module("." + module_name, __package__)
-    except Exception as e:
+    except Exception:
         print(f"!! Could not import PES module {module_name} !!")
         traceback.print_exc()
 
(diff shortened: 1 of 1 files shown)


ceriottm:
@DavideTisi  I think this could have helped you today.

ceriottm:
> So, just for my understanding: then if the driver cannot load the pes it will continue doing something anyways? Isn't that dangerous?

If it cannot load ONE PES, it will still be able to run with ANOTHER PES. Without this fix, if there's a problem importing say MACE, it won't be able to run with the harmonic PES - and it will not even tell you clearly it failed with loading MACE, just that it couldn't parse the pes module.

mahrossi:
ah I see, that is clear - I understand. but if it was asked to use in particular the PES that it could not load, then it will just refuse to do something.

ceriottm:
> ah I see, that is clear - I understand. but if it was asked to use in particular the PES that it could not load, then it will just refuse to do something.

yes it will give another error further down

ceriottm:
so can I merge this (and the ASE fix that I had branched from this one?)?

DavideTisi (approved):

DavideTisi:
for me you can merge

PR #362, opened 2024-08-07
Update README.md

ceriottm:


ceriottm:
Just a word change, seems more appropriate as description of a git clone installation

Issue #337, opened 2024-05-29
NVE simulations with `fixcom` use negative temperature when computing kinetic energy


svandenhaute:
When simulating in the NVE ensemble, in combination with the `fixcom` tag, the calculation of the kinetic energy still relies on `ensemble.temp`:
https://github.com/i-pi/i-pi/blob/68fb0b30c0ee4e4c5187aa64f96087e57ec7f227/ipi/engine/properties.py#L1094

Since this is set to `-1.0` when no temperature is specified (as one would expect), the square root evaluates to `nan`.

Potentially related to #336

mahrossi:
I think that instead we should enforce that a temperature is always set in case of nbeads > 1, even in the case of NVE - this is because if you run PIMD with NVE but no temperature, the spring constants make no sense. 

THEN for nbeads = 1 we really need to be more careful, because this hack to evaluate the temperature is not good anyways. Thanks for reporting.

ceriottm:
This should be fixed by #368

Issue #336, opened 2024-05-29
Unphysical Temperature Output in Partially-Adiabatic CMD Simulation


jorgecastro316:
I am encountering an issue with the output temperature of the centroid normal mode in my partially-adiabatic CMD. The simulation setup utilizes a two-temperature PILE-G thermostat. While the overall simulation runs as expected and the results seem physically reasonable, when the **fixcom** flag is set to **True**, the temperature of the centroid normal mode, as obtained by the **get_temp** function in **properties.py**, is unphysically high, which seems inconsistent with the rest of the simulation results.

venkatkapil24:
is this an issue only with partially adiabatic simulations? Can you check it for standard PIMD? What system are you running?

jorgecastro316:
No. It also happens for the centroid mode temperature  whenever I use PILE-G thermostat, in single or double temperature mode. I am running MAPI and Carboxilic Acid Fluoride

venkatkapil24:
it would be useful t have more info e.g. the centroid and higher normal mode temperatures (expected compared to observed).

mahrossi:
Hi @jorgecastro316 , could you check whether the problem persists after the latest cleanup? (latest version of i-PI). If so, can you share inputs/outputs from completeness?

mahrossi:
Related to #414 and to be addressed hopefully soon!

jorgecastro316:
Sadly the problem persist. I have been circumventing this issue and getting the correct temperature for the nm=0 if and only if i set the fixcom=False. Otherwise I get a value one order of magnitude larger than expected. This was done with the latest version of i-pi :/. Here are the inputs for a water monomer
[for_ipi.zip](https://github.com/user-attachments/files/18503754/for_ipi.zip)

mahrossi:
I think I will close this issue as it has been addressed with the latest devs by @venkatkapil24 and @weifang2321 .

@jorgecastro316 : can you please check that with the latest i-PI release you have none of these problems anymore? If any still remain, open a new issue with a full description, inputs and outputs. thanks!

PR #368, opened 2024-08-16
Clean up the kinetic temperature calculator

ceriottm:
Fixes errors when one (for whatever reason) prints temperature out of an NVE calculation


mahrossi (approved):
This looks fine to me and it gets rid of the NaN cases also flagged by #337 

Currently checking if it does anything to #336 but that should not stop this merge.

PR #369, opened 2024-08-18
A utility function to install the driver

ceriottm:
Until we figure out the added complexity of building wheels for the driver and include them in the pypi package, I figured that having an `ipi.install_driver()` utility function that will do all the work when running in an environment that has the basic build setup can be helpful. Matter of fact, I'm using this already for an [`atomistic-cookbook`](https://lab-cosmo.github.io/atomistic-cookbook/) example.


ceriottm:
maybe it's worth having the driver as part as the installation, but building wheels is beyond what I can achieve with chatgpt, so for the moment this would be it for me. thoughts?

ceriottm:
PSA: I'm going to keep adding unrelated fixes until this is merged ^_^

ipi/__init__.py
```diff
@@ -12,7 +13,7 @@
     "interfaces",
     "utils",
     "ipi_global_settings",
-    "read_output",
+    "install_driver" "read_output",
```
Luthaf, line 16:
```suggestion
    "install_driver",
    "read_output",
```

PR #375, opened 2024-09-21
Just cleaning up a bit the docstrings and the adaptive FD logic for the heat capacity estimators

ceriottm:
I've some doubts about the correctness of some of the S-C versions, but this should be the subject of a separate PR. This is mostly cosmetic (e.g. the docstrings of some of the estimators referred to the wrong papers).

### DIFF
--- a/ipi/engine/properties.py
+++ b/ipi/engine/properties.py
@@ -1855,8 +1855,8 @@ class Properties:
 
         return nx_tot / float(ncount)
 
-    def get_kscop_estimators(self, fd_delta=-_DEFAULT_FINDIFF):
-        """Calculates the op beta derivative of the centroid virial kinetic 
+    def get_scop_estimators(self, fd_delta=-_DEFAULT_FINDIFF):
+        """Calculates the op beta derivative of the centroid virial kinetic
         energy estimator for the Suzuki-Chin fourth-order path integral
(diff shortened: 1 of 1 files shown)


ipi/engine/properties.py
```diff
@@ -230,8 +230,8 @@ class Properties:
     """
 
     _DEFAULT_FINDIFF = 1e-4
-    _DEFAULT_FDERROR = 1e-6
-    _DEFAULT_MINFID = 1e-7
+    _DEFAULT_FDERROR = 1e-5
+    _DEFAULT_MINFID = 1e-8
```
mahrossi, line 234: All changes look good to me, but what is the driving force behind adjusting the defaults here?

ceriottm: Boh, I played around a bit with these defaults to try and find stabler ones. Not sure it helped TBH. Main realization is that you need VERY converged PPPM electrostatics to have stable finite difference calculations, but that's not i-PI's problem. Discuss this a bit in https://lab-cosmo.github.io/atomistic-cookbook/latest/examples/heat-capacity/heat-capacity.html

mahrossi: ok

mahrossi (approved):

PR #376, opened 2024-09-22
add xtb driver to python drivers

V-Alizade:
- implement xtb driver with tblite library
- add an example for the GFN2-xTB based on methane


mahrossi (commented):
This looks quite tidy to me! Thank you for the contribution. I would like to find a bit of time to try out the example in practice. Will set the automatic checks to run now.

drivers/py/pes/xtb.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy as np
+import json
+
+from ipi.utils.units import unit_to_internal, unit_to_user
+
+try:
+    import tblite.interface as tb
+except ImportError as e:
+    raise ModuleNotFoundError(
+        "Could not find tblite for xtb driver. Please install tblite-python with mamba"
```
mahrossi, line 18: This msg is not consistent with the pip installation that is suggested in the README (pip installation). What is best?

mahrossi:
I forgot about this - sorry -  I think it can be merged.

mahrossi (approved):

ceriottm:
@V-Alizade just a heads-up - we are revamping the PES infrastructure, so that it is also possible to run straight from i-PI without a driver (which is useful when running on a single node, and e.g. plain MD). 
in doing so I had to change the way the xtb driver works, by reading the config from a json file rather than a json formatted string on the command line. Comment on the #390  PR if you have issues on the tblite front with what I did.

V-Alizade:
@ceriottm I checked xtb client after your update, and everything looked fine. I also created an update to make it easier to use, please see https://github.com/i-pi/i-pi/pull/407

PR #380, opened 2024-09-24
Better dev interface to force.copy (now renamed force.clone)

ceriottm:
Forces.copy allowed to pass None for the target beads and cell object, but it's likely this was always broken.
To make life easier for devs, beads and cell are now mandatory, docstring is a bit clearer, and function is renamed to clone() to hint that it does more than a shallow (or deep) copy.

### DIFF
--- a/examples/features/geometry_optimization/BFGS_pswater/input.xml
+++ b/examples/features/geometry_optimization/BFGS_pswater/input.xml
@@ -1,4 +1,4 @@
-<simulation mode="static" verbosity="high">
+<simulation mode="static" verbosity="medium">
    <output prefix='simulation'>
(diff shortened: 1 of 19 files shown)


mahrossi (approved):
Yes, I also tested a simpler version of this change locally and it does what it should - allowing future devs not to waste hours looking for the strangest possible behaviour of a "copy" call. Great that it is not called copy anymore.

mahrossi:
Ok, tested locally - explanation coming in the merge msg I think. Let's merge this. I left some comments to, in the future, get rid of some transfer_force_manual calls.

mahrossi (approved):

PR #382, opened 2024-10-03
Added a utils library for ACF and wrapped the postprocessing tool around it

ceriottm:
Thought it would be good to put some order in the post-processing tools, and also that it was a good point to start exploring the process of making i-PI more "library-like"


mahrossi:
Yes, I take it that for now, in the way you changed it, calling the script as one did before yields the same result, right? Only the core has been "librarized". Looks good.

ceriottm:
> Yes, I take it that for now, in the way you changed it, calling the script as one did before yields the same result, right? Only the core has been "librarized". Looks good.

Yes, nothing changed from a "user" perspective (also because I vaguely recall there are tutorials that use this).

Issue #379, opened 2024-09-24
Bug report: instools.py: get_imvector(h,m3)


Phyzch:
Hi, there,
    I find there is a bug in instools.py : get_imvector(h,m3) which is used to extend the initial instanton path around the transition state. 

In line 212, the code is : imv = w[:, 0] * (m3[:] ** 0.5) 

I think this is wrong, because w[:,0] is the eigenvector of mass scaled coordinate \sqrt{m} q, therefore, to get the eigenvector of the coordinate q, we should have:  
imv = w[:,0] / (m3[:] ** 0.5).

I test the code use my own example, where the code fails to give the correct unstable mode. After I correct it, it gives the right unstable mode. Please let me know what you think. Thank you.

Best,
Chenghao Zhang.

litman90:
Hi Chenghao,
thanks for reaching out.
I will check this next week and let you know.
Best,
Yair

litman90:
Hi Chenghao,
I apologize for the slow response. I hade a closer look at your comment and your observation is right. Could you create a pull request with the bugfix?
Thanks for your contribution and for reaching out
Best,
yair

litman90:
merged. Thanks!

PR #383, opened 2024-10-08
bugfix instools

litman90:
1 line bugfix reported here #379


litman90:
It is a 1 line bugfix. It's not worth bothering other maintainers.

PR #384, opened 2024-10-13
Moved GLE spectrum deconvolution code to tools

ceriottm:
One more tool script restructured to facilitate usage from python


ceriottm:
Harmless enough that  I'll merge - needed for cookbook.

PR #389, opened 2024-11-02
Prevent new `xtb` module from crashing simulations

frostedoyster:
If `tblite` is not installed, an import is attempted and this crashes i-PI due to the error raise. This should fix it


ceriottm:
I'm puzzled as to how you stumbled onto this, but thanks for the fix!

ceriottm (approved):
lgtm

PR #390, opened 2024-11-10
A "direct" FF using the same PES interfaces as the python driver

ceriottm:
This PR implements a "direct" FF that avoids the entire socket communication and allows to directly call the PES interfaces that were previously implemented for the python driver only.


litman90:
@ceriottm I fixed pes examples and regtests related to the DW pes, 
I let you fix the remaining two examples which are  failing

test folder: /home/runner/work/i-pi/i-pi/examples/features/input_output/replay
test folder: /home/runner/work/i-pi/i-pi/examples/features/replica_exchange/pi_remd_liquid_water

PR #394, opened 2024-11-11
Implement fixatoms_dof tag and variables

litman90:
Draft pull request


litman90:
@mahrossi, could you review it?
This is the first part of extending fixatoms to fixdof to treat 'simple' constrain in a more general way.
at the moment there are two tags fixatoms  and fixatoms_dof, but internally we have only the fixatoms_dof variable everywhere

In a next PR, we would like to merge fixatoms and fixdof tags and use a syntax like
<fixatoms mode='x'>[...] </fixatoms> 
For this, a new input class needs to be created.

litman90:
@mahrossi, I've added examples and regression tests.

mahrossi:
I will merge this

PR #400, opened 2024-11-12
Gitinfo

EliaStocco:
Created function that add information (git branch, last commit, machine name, username, etc) that are printed in the log file, properties file, and checkpoints


ipi/utils/messages.py
```diff
@@ -96,6 +104,190 @@ def __setattr__(self, name, value):
...
+    else:
+        info_string += "# Unable to retrieve Git information.\n"
+
+    info_string += "#\n"
+
+    if system_info:
+        info_string += "# System information:\n"
+        info_string += f"#     Current Folder: {system_info['current_folder']}\n"
+        info_string += f"#       Machine Name: {system_info['machine_name']}\n"
+        info_string += f"#               FQDN: {system_info['fqdn']}\n"
```
mahrossi, line 279: no need  for FQDN I think

ipi/utils/messages.py
```diff
@@ -96,6 +104,190 @@ def __setattr__(self, name, value):
...
+        info_string += f"#   Operating System: {system_info['os_name']}\n"
+        info_string += f"#         OS Version: {system_info['os_version']}\n"
+        info_string += f"#          Processor: {system_info['processor']}\n"
+        info_string += f"#     Number of CPUs: {system_info['num_nodes']}\n"
+        info_string += f"#          User Name: {system_info['user_name']}\n"
```
mahrossi, line 284: also no need for user name? I am unsure about the necessity of some other fields, but ok, why not!

mahrossi (changes requested):
Very nice in my opinion. Just a few very small comments. Will be an important add on to enhance reproducibility.

litman90:
We had a discussion online with Elia and Michele and decided on the following:

- [ ]  In the RESTART file and log.out file, we print the verbose version
- [ ]  In the properties, we add an 'opt-in' flag, the default is not to print the info
- [ ]  Elia needs to check on not breaking the internal parser (and fix it if necessary)
- [ ]  if one opts in to output the information in the properties file, it should be done only once

EliaStocco:
The information messages are always printed to the log file and checkpoints/RESTART files.
I have added a `verbosity` attribute to `PropertyOutput`, which can assume only low and high values:
- `low`: usual behavior
- `high`: print the information messages to the properties output file too starting with `###`

I have also added two `pytest` tests to check that read_output works for both the version of the outputs (with verbosity high or low). It actually worked out of the box and no change to the code was required.

mahrossi:
This looks ready - are there any issues we need to check @EliaStocco ?

ceriottm (commented):
Thanks a lot, this is a great idea. Unfortunately there are _many_ issues that need being addressed. On the macro level: 
(1) it looks like you branched from the consolidated-ASE branch so that needs to be merged first, then this rebased on the new main
(2) the way tracking of the version is done is not super-useful, as it works only if i-PI is executed from a git version, without installation. this is a corner case that applies well basically to us core devs. most users will pip install from pipy, or at most from a branch - so there should be a mechanism to also incorporate the version in the installed source so that it can be printed
(3) I know we discussed this, but seeing the "optional output" implemented really feels uber-clunky. how about we agree on having a ONE-LINE version summary in the header, e.g. pointing at the commit, and we make it part of the standard after checking it doesn't break the parser?
(4) You can (and probably should) wrap the whole version info in a single XML comment, rather than having one per line. 
Can also discuss this on slack, if it's easier and faster.

mahrossi:
1) because the branches are mingled now, it is better to close this pr and start another one with the gitinfo modifications.

2) hopefully git info applies not only to core devs but to any developer

3) to me, how exactly it is outputted to the files is a matter of taste, and many lines vs. one line is really something that makes very little difference

4) we should agree on what we want to output if the user is using the installed version. only the release version number, yes?

ceriottm:
> 2. hopefully git info applies not only to core devs but to any developer
The problem is not that it prints the git info, but that it only prints the version if you are running i-PI from the git repo folder. If you pip install it it has no way of getting the version right now.

PR #402, opened 2024-11-13
Update docs

litman90:
This PR adds a diagram to explain the communication protocol.


docs/figures/ipi-comm.svg
mahrossi, line 1: nice

mahrossi: it is a small thing but maybe "evaluate forces"  can be made more generic. anyways, for the good reader the generality is understood

mahrossi: also the box "demands energy and forces to continue" - the meaning is better portrayed as "needs new energy and forces to continue"

litman90: modified as requested

litman90:
@mahrossi  it should be good to go now.

mahrossi (approved):

PR #405, opened 2024-11-13
Fix multiple random number outputs

mahrossi:
@venkatkapil24 have a look.


ipi/inputs/simulation.py
```diff
@@ -57,7 +57,7 @@ class InputSimulation(Input):
             InputRandom,
             {
                 "help": InputRandom.default_help,
-                "default": input_default(factory=Random),
+                "default": input_default(factory=(lambda: Random())),
```
ceriottm, line 60: I don't think this is necessary if you don't hardcode a default seed
```suggestion
"default": input_default(factory=Random),
```

ipi/engine/simulation.py
```diff
@@ -161,6 +162,12 @@ def __init__(
         self.syslist = syslist
         for s in syslist:
             s.prng = self.prng  # bind the system's prng to self prng
+            if threading.current_thread() == threading.main_thread():
+                info(
+                    "@ RANDOM SEED: The seed used in this calculation was "
+                    + str(self.prng.seed),
+                    verbosity.low,
+                )
```
ceriottm, line 170: nice!

ceriottm (changes requested):

ipi/inputs/simulation.py
mahrossi: I forgot to erase the lambda:

mahrossi (commented):
done and ready to merge

PR #406, opened 2024-11-14
Elphmod

EliaStocco:
I have tidied up the code


mahrossi:
Yes, this looks better.

ipi/pes/elphmod.py
```diff
@@ -14,18 +14,16 @@ class ModelIIIDriver(Dummy_driver):
     Example: python3 driver.py -u -m elphmod -o driver=driver.pickle
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, driver, *args, **kwargs):
```
mahrossi, line 17: yep, I think this is the right thing to do.

mahrossi (approved):

PR #407, opened 2024-11-16
update xtb client

V-Alizade:
- make xtb client more user-friendly by reading .xyz input
- create xtb calculator once

### DIFF
--- a/examples/clients/xtb/README.md
+++ b/examples/clients/xtb/README.md
@@ -15,5 +15,5 @@ the example can be run as usual
 
 ```bash
 i-pi input.xml &> log.ipi &
-i-pi-py_driver -m xtb -o '{"method": "GFN2-xTB", "numbers": [6,1,1,1,1], "periodic": false}' -u -a xtb &> log.xtb &
+i-pi-py_driver -m xtb -o template=ch4.xyz,method=GFN2-xTB -u -a xtb &> log.xtb
 ```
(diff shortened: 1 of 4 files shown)


ceriottm (approved):
Thanks for this PR. I've one main comment, but I leave this for you to decide if it matters so if you prefer to go ahead with this, I approve the PR already. 
The way you implement the reading of the xyz file assumes this is in "i-PI xyz" format. I am not sure this would be super-friendly to xtb users (with hindsight, it was a bad decision to make our own format) and since this is the driver-side input you may as well require that it is say extxyz formatted and read it with ase.io.read. As an additional bonus, you can then get the periodic flags from there, and the atomic numbers are automatically inferred so you don't need to add the whole dictionary.

V-Alizade:
Thank you for your comment. Let's keep it as it is. The tblite library (GFNn-xTB method) is already interfaced with DFTB+ and ASE, making it available through these clients as a calculator. However, having it as a standalone client might improve performance and reduce dependencies.

Additionally, the elements dictionary is no longer needed with the new tblite release (v0.4). I will update it accordingly.

V-Alizade:
> Cool, then I let you update the code to be compatible with this new release, and then merge!

Please have a look, I updated for the newest version.

ceriottm (approved):
I didn't manage to install xtb so I couldn't test it, but if it works on your end, LGTM

PR #409, opened 2024-11-21
Fix traj parsing

ceriottm:
@EliaStocco  let's fix here the parser, before moving to the consolidation mechanism.

### DIFF
--- a/ipi/utils/parsing.py
+++ b/ipi/utils/parsing.py
@@ -139,7 +139,6 @@ def read_trajectory(
             raise ValueError(f"Unrecognized file format: {format}")
 
     file_handle = open(filename, "r")
-    bohr2angstrom = unit_to_user("length", "angstrom", 1.0)
     comment_regex = re.compile(r"(\w+)\{([^}]+)\}")
     step_regex = re.compile(r"Step:\s+(\d+)")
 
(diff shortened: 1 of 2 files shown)


ceriottm:
given that it looks like we've kind of lost momentum following the hackaton, I'd like to merge this, given that the current code is buggy. I would also then make a point release so we can start using and improving the scripting interface, ffdirect, the atom constraining, and all the other nice add-ons that were implemented.

ipi/utils/parsing.py
```diff
@@ -219,13 +219,37 @@ def read_trajectory(
...
+                    frame.positions *= 0.0  # trajectory does NOT contain position data!
+                    frame.arrays[what] = ret["atoms"].q.reshape((-1, 3))
+                    if what in [
+                        "forces",
+                        "forces_spring",
+                        "forces_component",
+                        "forces_sc",
+                        "Eforces",
+                        "f_centroid",
+                        "forces_component_raw",
```
mahrossi, line 241: so every time we add a new position keyword of force keyword we have to hand-modify this list, yes? maybe worth adding a disclaimer about this in the file.

mahrossi (approved):

ipi/utils/parsing.py
ceriottm: Yes, I don't see an easy way around it unfortunately.

ceriottm: Actually I didn't think hard enough. Thanks for spotting this, there as indeed a way to avoid the manual mess.

mahrossi (approved):

mahrossi:
Amazing, this looks way better. I will merge.

PR #411, opened 2024-12-03
Improve the PLUMED interface

ceriottm:
This changes a few things in how the PLUMED interface works, so that it does fewer unnecessary colvar calculations (useful for expensive CVs) and provide a mechanism to extract the CV values into an extras field. 
Might also add a few extra examples.


robinzyb:
[benchmark-opes-ipi.tgz](https://github.com/user-attachments/files/18086289/benchmark-opes-ipi.tgz)
I tested the OPES-i-PI simulations. The results look good. Please check out the`benchmark-opes.ipynb`. The only problem is that the `*.colvar_0 `files contain only zeros. Is this expected? or a bug?

robinzyb:
the problem solved. I tested with latest plumed in github.
Screenshot 2024-12-12 at 15 19 02
Screenshot 2024-12-12 at 15 19 13
CVs can be captured by i-PI

ceriottm:
> Just for memo, benchmarked results with normal metad beads=8 and PACE=50.

Thanks for checking - and good to see that you could easily adapt the plotting scripts from [the cookbook recipe](https://atomistic-cookbook.org/examples/pi-metad/pi-metad.html)

I'll see if I can get a review from another maintainer, otherwise I'll merge and maybe make a point release since we've quite a few new bugfixes and features that got in lately.

DavideTisi (approved):

ceriottm:
Also fixes the problem mentioned in #412

PR #412, opened 2024-12-26
fix CI

DavideTisi:
Last commit broke the CI because of a change in the setup script, which pulled a non existing branch


ceriottm:
Problem should be fixed already by #411

PR #413, opened 2024-12-28
Release/v3.1

ceriottm:
3.1 release, containing multiple fixes and new features such as a direct Python forcefield mode, and scriptable i-PI runner.

PR #415, opened 2025-01-20
update python version in setup.cfg

litman90:
@ceriottm 
1 line PR.
 PR #413 didn't update the Python version to 3.8 in the setup.cfg file.


ceriottm:
Strange I think I did that. There is also a stray regtest CI item that doesn't run, I'll use this PR to try and fix that as well

ceriottm:
Is this is?

PR #417, opened 2025-01-20
Bugfix/get temp

venkatkapil24:
tests needed


venkatkapil24:
@weifang2321 regression_tests/tests/TEMPERATURE/nvt_md_fixatoms/ something wrong in temperature output with fixatoms

weifang2321:
> @weifang2321 regression_tests/tests/TEMPERATURE/nvt_md_fixatoms/ something wrong in temperature output with fixatoms

The problem is that ipi sets fixcom=True by default, and that fixatom conflicts with fixcom. We should make the two options mutually exclusive

venkatkapil24:
I may have found the issue. It is in dynamics/thermostat. [Line 167 of dynamics.py](https://github.com/venkatkapil24/i-pi/blob/main/ipi/engine/motion/dynamics.py#L167) defines an effective number of dofs for the thermostat, however, this happens before the [override](https://github.com/venkatkapil24/i-pi/blob/main/ipi/inputs/motion/motion.py#L323). 
@ceriottm do you have a suggestion for a clean rearrangement? This may also affect global thermostats with constraints.

weifang2321:
> I may have found the issue. It is in dynamics/thermostat. [Line 167 of dynamics.py](https://github.com/venkatkapil24/i-pi/blob/main/ipi/engine/motion/dynamics.py#L167) defines an effective number of dofs for the thermostat, however, this happens before the [override](https://github.com/venkatkapil24/i-pi/blob/main/ipi/inputs/motion/motion.py#L323). @ceriottm do you have a suggestion for a clean rearrangement? This may also affect global thermostats with constraints.

For the thermostats it seems that ndof only affect SVR and CL

venkatkapil24:
Wei is right; the of variable doesn't affect Langevin dynamics, so it's not the source of the problem. Turns out however, the fixcom in dynamics is never overwritten. This is the issue. So the constraints class gets fixcom True even when fixatoms is a finite list.

venkatkapil24:
the issue is fixed

venkatkapil24:
There is still an error in the temperatures for partially adiabatic centroid molecular dynamics

weifang2321:
setting nm mode to pc-cmd does not modify the m3 in the beads object, therefore obviously the bead KE calculated with get_kinmd would be lower than the ture KE due to wrong mass. I don't think the bead temperature is well defined in this case therefore we should not worry about it and should add a warning for people trying to get bead temperature in pacmd

ceriottm:
> setting nm mode to pc-cmd does not modify the m3 in the beads object, therefore obviously the bead KE calculated with get_kinmd would be lower than the ture KE due to wrong mass. I don't think the bead temperature is well defined in this case therefore we should not worry about it and should add a warning for people trying to get bead temperature in pacmd

So, in the NM basis the temperature IS well-defined, so I wouldn't brush off entirely the possibility it can be defined also in the beads representation. Worth a thought. If not, then I think it'd be great to be able to issue a warning if someone prints the bead temperature when running PA-CMD.

ceriottm (approved):
Cool. I tested, cleaned up the code a tiny bit, and it all looks good to me. Great to have removed the crazy "add and remove velocity" dance we were doing. Didn't realize there was such a simple closed-form expression for the COM energy contribution.

PR #419, opened 2025-01-20
Input versioning

yotamfe:
Hi,

Adding a version field to the simulation. As a soft introduction of the feature, the code issues a warning when no version is provided, and an error if the version is not recognized. Also added a script that edits an existing input file and adds the version field.

Deferring changes to all the examples in the repository until we agree on the feature's design.

Thanks!


yotamfe:
Also, changed where the i-Pi version is placed - in a python file and read from the config, rather than in the config and obtaining it in the code through importlib.metadata. The reason is that the latter only works if the code was installed. I didn't end up referencing the i-Pi version for input versioning (in order to keep these two entities decoupled), but keeping this change here for now in order to have the option to do that.

I tested the current scheme with a direct installation (python setup.py install).

yotamfe:
Hi @mahrossi,

Following the discussion in the last meeting, I'm waiting with updating the input files in all the tests and marking the PR ready for a light review of the functionality design choices - when a warning is issued and when an error. Let me know what you think.

Thanks for considering this contribution!

mahrossi:
Hi there, I think we discussed this in the last meeting already and concluded what to do - yes, @yotamfe ? If so, then let me know when this PR is not a draft anymore.

yotamfe:
Hi @mahrossi 

Sorry, missed this message while traveling. I've tested reading the version number and running a tiny simulation (1) a non-installed version, (2) with an installation using setup.py, (3) with an installed package (from `python -m build; pip install dist/<package_name>.whl`), (4) a non-installed version with a different version number than the installed package.

Two questions:

(1) As far as I understand, the directive in the file [pyproject.toml](https://github.com/i-pi/i-pi/blob/main/pyproject.toml)
`
[tool.setuptools_scm]
write_to = "ipi/_version.py"
`
means that package generation ignores the value for the version in [setup.cfg](https://github.com/i-pi/i-pi/blob/main/setup.cfg) and instead uses information from git. So for instance the package generated now is ipi-3.1.2 (which is the latest release) even though the value in [setup.cfg](https://github.com/i-pi/i-pi/blob/main/setup.cfg) is 3.1. Do I understand correctly? Is this intentional?

(2) At this point I only modified one input file in the examples directory to use the version number (see [this commit](https://github.com/i-pi/i-pi/pull/419/commits/4499cb0d02f38f42f72dfc494822546a83674e29)). I used the [script](https://github.com/yotamfe/i-pi/blob/input-versioning/tools/py/input_file_add_version_3.1.py) supplied for this purpose. I can do this for all examples, but the changes are not local (because the script parses the XML and regenerates it as text). The disadvantage is clear, the advantage is perhaps a sort of standardization of all the inputs. Another consideration is that the value would need to be modified in all examples that include it whenever we change the version number, although this shouldn't be too bad if we supply such a script every time we do this (which is good practice in my opinion).

Please let me know what you think and I'll mark the PR as ready for review.

Thanks for considering this contribution!
Yotam

yotamfe:
Thanks @mahrossi! Marking the PR as ready for review.

I removed the directive from pyproject.toml, so now when the package is generated the version is taken from ipi/__init__.py rather than the tag name. The version number appears in the code only once.

I tested the code when i-Pi is not installed, when it's installed as a (local) pip package, and when installed through setup.py.

If the version number is missing, a warning is issued, but the simulation will run. We will remember to update the example because it will fail the test - if a version number is present and incorrect, it's an error, not a warning.

The RESTART and checkpoint files already have the i-Pi version.

Thanks for looking into it!

mahrossi (commented):
I iterated this with @yotamfe and I am happy with it, so I can approve the commit. I would be happier if @ceriottm had a quick look just because it does add a bit of something new in the code here that will stop future simulations with different code versions restarting from old RESTART files. anyways I think it is agood move for code maintenance.

mahrossi:
I am closing this as it got stale but I mention it in a newer PR, above.

Issue #422, opened 2025-02-19
Conserved quantity drift with constrained atoms


mahrossi:
The conserved quantity drifts when constraining atoms. Below is a plot for the water monomer with the pswater potential. No constraints and center of mass constraint are fine, but one and two atoms show the drift. Happens for NVT or NVE, already for classical simulations (1 bead). For NVE comparing the conserved quantity and just potential+kinetic, the conserved quantity shows the linear drift, while potential+kinetic is conserved, as expected.

There is some issue related to how the kinetic energy and `econs` are calculated (check `ensembles.py` in `get_econs` and `motion/dynamics.py` in `pconstraints`).  

The drift gets smaller with smaller time steps but slope increases with the number of fixed atoms. I started debugging this yesterday but could not yet find an obvious way of fixing it. Person who found this is @krystofbrezina 

[inputs.zip](https://github.com/user-attachments/files/18862669/inputs.zip)

![Image](https://github.com/user-attachments/assets/1c31c752-0a28-48b4-a4be-9d94b482c574)

mahrossi:
There are some fixes going on in #423

PR #423, opened 2025-02-19
Removing drift from conserved quantity.

mahrossi:
Starting to address the problems reported in #422 . Need to check and discuss whether this fix is general.
In the NVE step I propagate the momenta only of the active dofs (not of the fixed ones). 

This is in draft stage, there are debugging print statements left. Still did not check many simulation types and combinations.

### DIFF
--- a/ipi/engine/motion/dynamics.py
+++ b/ipi/engine/motion/dynamics.py
@@ -221,6 +221,11 @@ class Dynamics(Motion):
             elif self.enstype == "nst":
                 if np.allclose(self.ensemble.stressext.diagonal(), -12345):
                     raise ValueError("Unspecified stress for a constant-s integrator")
+        if self.enstype == "nve" and self.beads.nbeads > 1:
+            if self.ensemble.temp < 0:
+                raise ValueError(
+                    "You need to provide a positive value for temperature inside ensemble to run a PIMD simulation, even when choosing NVE propagation."
+                )
 
     def get_ntemp(self):
         """Returns the PI simulation temperature (P times the physical T)."""
--- a/ipi/engine/motion/dynamics.py
+++ b/ipi/engine/motion/dynamics.py
@@ -288,8 +288,12 @@ class DummyIntegrator:
         self._dt = motion._dt
         self._nmts = motion._nmts
 
-        # calculate active atoms
-        full_indices = np.arange(len(self.beads.p[0]))
+        # check whether fixed indexes make sense
+        if np.any(self.fixatoms_dof >= (3*self.beads.natoms)):   
+            raise ValueError("Constrained indexes are out of bounds wrt. number of atoms.")
+
+        # calculate active dofs
+        full_indices = np.arange(3*self.beads.natoms)
         self.activeatoms_dof = np.setdiff1d(full_indices, self.fixatoms_dof)
 
         # total number of iteration in the inner-most MTS loop
--- a/ipi/engine/motion/dynamics.py
+++ b/ipi/engine/motion/dynamics.py
@@ -296,7 +296,10 @@ class DummyIntegrator:
 
         # calculate active dofs
         full_indices = np.arange(3 * self.beads.natoms)
-        self.activeatoms_dof = np.setdiff1d(full_indices, self.fixatoms_dof)
+        # in the next line, we check whether the fixed indexes are in the full indexes and invert the boolean result with the tilde
+        mask = ~np.isin(full_indices, self.fixatoms_dof)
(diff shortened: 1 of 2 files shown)


ipi/engine/motion/dynamics.py
```diff
@@ -361,12 +365,11 @@ def pconstraints(self):
...
-            self.ensemble.eens += np.sum(vcom**2) * 0.5 * Mnb  # COM kinetic energy
+            self.ensemble.eens += np.sum(vcom**2) * 0.5 * Mnb  # COM kinetic energy.
 
         if len(self.fixatoms_dof) > 0:
             m3 = dstrip(beads.m3)
             p = dstrip(beads.p)
-
             self.ensemble.eens += 0.5 * np.sum(
                 p[:, self.fixatoms_dof] ** 2 / m3[:, self.fixatoms_dof]
             )
```
litman90, line 375: here we are setting the the momenta of the fix dof to zero, 
The changes in line 411ish mean that the momenta of those dof is not updated.

I might be missing something, but I don't understand why we need to do both things.
Do these changes fix the drift you observe?

ipi/engine/motion/dynamics.py
```diff
@@ -283,6 +283,10 @@ def bind(self, motion):
         self._dt = motion._dt
         self._nmts = motion._nmts
 
+        # calculate active atoms
+        full_indices = np.arange(len(self.beads.p[0]))
+        self.activeatoms_dof = np.setdiff1d(full_indices, self.fixatoms_dof)
+
```
litman90, line 289: I think this place to get the active dof is ok for the reasons you described.
Indeed, I would suggest adding a check here to verify if the indices are compatible with the number of atoms (i.e. np.all(self.fixatoms_dof <= 3*self.natoms) )

litman90:
@mahrossi, I am taking 2 weeks off starting today so I won't be able to review further changes or be active in the discussion of this PR in the next two weeks, (Indeed, I am reviewing this from the airport).  Feel free to assign another reviewer if you want this to get done before I come back. I apologize for the inconvenience.

mahrossi:
> @mahrossi, I am taking 2 weeks off starting today so I won't be able to review further changes or be active in the discussion of this PR in the next two weeks, (Indeed, I am reviewing this from the airport). Feel free to assign another reviewer if you want this to get done before I come back. I apologize for the inconvenience.

no worries, Krystof is stepping in and others will help. we will merge this ASAP.

krystofbrezina:
I have tested the new fix on the case of constrained water dynamics using the i-PI driver `pswater` potential. Indeed, it seems that the drift of the conserved quantity is fixed, while the dynamics (from looking at the evolution of the potential over the trajectories) is not changed. In the legends, `w <ensemble>` means that for the NVE runs, the section defining the ensemble temperature `<ensemble><temperature units='kelvin'>300.0</temperature></ensemble>` is present, in the case without the label it is not.
[conserved.pdf](https://github.com/user-attachments/files/18905472/conserved.pdf)
[potential.pdf](https://github.com/user-attachments/files/18905474/potential.pdf)

ceriottm:
MAYBE rather than using a setdiff we can create masks? IDK maybe it will be faster / easier given that neither fixdof nor activedof need to be sorted. Also let's check this doesn't slow down everything when running without constraints.

mahrossi:
Could you suggest which systems to test for speed - do we have a test set e.g. with increasing number of atoms? Masks ok if they are really much faster. The geop "branch" of i-PI works with masks and I find it harder to follow. I can pick this back up in the hackathon - this week I can't put time here.

ceriottm:
> Could you suggest which systems to test for speed - do we have a test set e.g. with increasing number of atoms? Masks ok if they are really much faster. The geop "branch" of i-PI works with masks and I find it harder to follow. I can pick this back up in the hackathon - this week I can't put time here.

I'm thinking the profiling examples that are in the tests folder. Those with minimal extra stuff like plain NVE with no forcefield.

mahrossi:
hey @ceriottm I switched to direct boolean masks, which should indeed be faster. I see a small overhead created by this when running a test I made based on the classical gas profiling example. It is on the order of 1e-4 s/step for the examples I tested. It is a bit hard to quantify further because my computer keeps running slightly different times every time I run. I tested up to 1024 particles with 10 dofs fixed. I can test more if you want, but it does not seem to be major. Let me know if it is ok to merge for you.

mahrossi:
@ceriottm or @litman90 please merge this if all is fine.

ceriottm (approved):
LGTM. Let's hope it does not introduce side-effects elsewhere

PR #424, opened 2025-02-21
Implement a native `metatensor` driver

frostedoyster:
Compared to the old driver, this driver can return energy, force and virial ensembles


DavideTisi:
so checking the example in `ipi/example` with Nikel the result with version are different wrt using ipi main branch

ceriottm:
Is this ready to check?

DavideTisi:
The code works. 
It would be nice to have a test, comparing maybe with ase. The ensemble was not tested at all.

ceriottm (approved):
OK after some iterations with @frostedoyster I think this is good to go. We also added an example. We'll test this more thoroughly when we write a tutorial for UQ with PET-MAD

PR #425, opened 2025-02-27
Batch ffdirect

EliaStocco:
This forks aims to find a way to let `i-PI` send many configurations (from different beads) to the same ML model.
This might be extremely helpful when evaluating the ML model on GPU because the ML model can evaluate many different configurations at the same time better exploiting GPU.
For optimization purposes, `FFDirect` is exploited, by avoiding `socket` communication.
For the moment, I am focusing on `MACE` only.
Only the code necessary to i-PI is provided, while the code necessary to the ML model to read all the configurations will be provided soon. The latter however can be customized by the user.

### DIFF
--- a/ipi/engine/forcefields.py
+++ b/ipi/engine/forcefields.py
@@ -383,13 +383,13 @@ class FFEval(ForceField):
 
         # We have to be thread-safe, as in multi-system mode this might get
         # called by many threads at once.
-        with self._threadlock:
-            for r in self.requests:
-                if r["status"] == "Queued":
-                    r["status"] = "Running"
-                    r["t_dispatched"] = time.time()
-                    self.evaluate(r)
-                    r["result"][0] -= self.offset  # subtract constant offset
+        # with self._threadlock:
+        for r in self.requests:
+            if r["status"] == "Queued":
+                r["status"] = "Running"
+                r["t_dispatched"] = time.time()
+                self.evaluate(r)
+                r["result"][0] -= self.offset  # subtract constant offset
 
     def evaluate(self, request):
         request["result"] = [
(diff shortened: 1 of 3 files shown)


ipi/engine/forcefields.py
```diff
@@ -383,13 +383,13 @@ def poll(self):
...
         # We have to be thread-safe, as in multi-system mode this might get
         # called by many threads at once.
-        with self._threadlock:
-            for r in self.requests:
-                if r["status"] == "Queued":
-                    r["status"] = "Running"
-                    r["t_dispatched"] = time.time()
-                    self.evaluate(r)
-                    r["result"][0] -= self.offset  # subtract constant offset
+        # with self._threadlock:
```
mahrossi, line 386: hey Elia, I think we can't safely uncomment this. could you explain your rationale here, so we could find another solution?

ipi/pes/fileIOCalculator.py
```diff
@@ -0,0 +1,334 @@
...
+import re
+import numpy as np
+from datetime import datetime
+from filelock import FileLock
+from ase import Atoms
+from ase.io import write
+from ase.calculators.calculator import Calculator, all_changes
+from typing import Dict, Union, Any
+from .ase import ASEDriver
+
```
mahrossi, line 14: here make sure to use as few modules as you can in order to decrease dependencies and make code more portable.

ipi/pes/fileIOCalculator.py
```diff
@@ -0,0 +1,334 @@
+import os
```
mahrossi, line 1: could you explain the intended purpose of this routine? I am not sure I am getting why so much code is needed, but I think I just need some explanation :-)

EliaStocco:
@mahrossi @ceriottm Shall I close this PR since these changes will not be merged?

PR #426, opened 2025-03-03
PSwater pytorch

EliaStocco:
I have implemented PSwater as a driver using `pytorch` so that I can also have the spatial derivatives of the dipole (Born charges) and use them for debugging the implementation in i-PI that deals with time external electric fields.


ceriottm:
maybe you can wrap it with metatrain so it's the same interface as [this example](https://atomistic-cookbook.org/examples/water-model/water-model.html)

ipi/pes/ase.py
mahrossi, line 1: I think this file is also trying to be merged in another commit. But ok, at max it will generate conflict

mahrossi:
Hi @EliaStocco, making pswater torch-mented is a great idea and a great contribution. We should make it happen. I think right now it is a bit convoluted. I looked at the cookbook that @ceriottm suggested and I have a few questions regarding maintainability. In any case we need to simplify a bit the current implementation and make the tests pass. Good topic for tomorrow.

EliaStocco:
I will use `metatensor` to provide a `pytorch` implementation of `PSwater`

PR #429, opened 2025-03-07
Bug in FFDirect

EliaStocco:
**Problem**: 
The extra strings are not properly parsed in `FFDirect`.
Actually they are just not parsed at all.

**Solution**: 
I have just copied and pasted the code from `Driver.getforce` in `ipi/interfaces/sockets.py` to `FFdirect.evaluate` in `ipi/engine/forcefields.py`.


.github/workflows/documentation.yml
```diff
@@ -1,11 +1,6 @@
 name: Documentation
 
-on:
-  push:
-    branches: [main]
-    tags: ["*"]
-  pull_request:
-    # Check all PR
+on: [push, pull_request]
```
mahrossi, line 3: clean up changes you did to the workflows files. none of them should be there.

.github/workflows/examples.yml
```diff
@@ -1,10 +1,6 @@
 name: Pytest for examples
 
-on: 
-  push:
-    branches: [main]
-  pull_request:
-    # Check all PR
+on: [push, pull_request]
```
mahrossi, line 3: I think this was accidental, remove. Also all others.

PR #432, opened 2025-03-18
Feat/mace

EliaStocco:
I found some issues using MACE via the provided i-PI driver.
I think the problem was in the way properties were handled in the ASE driver.
I made some minor changes that should not alter the behavior of other drivers but allow other properties to be provided in addition to energy, forces and stress.


ipi/pes/ase.py
```diff
@@ -81,15 +82,27 @@ def __call__(self, cell, pos):
         structure.calc = self.ase_calculator
 
         # Do the actual calculation
-        properties = structure.get_properties(self.capabilities)
+        self.ase_calculator.calculate(structure)
```
Luthaf, line 85: Depending on the calculator, one might still have to explicitly pass the requested properties to compute; so we might want to keep the previous way of calling this, but adding more stuff to `self.capabilities`, according to user request.

frostedoyster:
I agree that, in principle, asking for all properties can be wasteful in an arbitrary model. In this case, since we know the architecture of MACE, I would say that:
- asking for forces and stresses when only the energy would be needed is a theoretical overhead of around 100% (2x the energy cost). I think i-PI always wants forces and stresses though
- asking for equivariant properties when only energies, forces and stresses are needed should be, roughly, a <20% overhead on the forward (you waste some computation by calculating equivariants in the last layer when you don't need them), and therefore a <10% overhead overall

frostedoyster:
Ah, I didn't realize that! I agree with you @Luthaf

mahrossi:
@EliaStocco could you address the changes?

mahrossi:
I looked at the code and I am not sure how to achieve this simply only by modifying the MACE driver. I think Elia had two problems, but he was trying to solve generally.

I imagine that @Luthaf is right and what we need is a way to let the user get more properties at will. We can make some flags such as "all" (everything that the client will return), "standard" (energy, forces, stresses), and also accept a manual list with any entries the user wants. What I don't understand is what was the problem that @EliaStocco was having with the call as it was at first, as it seemed a standard call to "get_properties". A short explanation could give us a best route here.

ceriottm:
Maybe @EliaStocco you can have a look at the metatensor driver in #435 . This is a very similar problem (although in this case we want to compute a different type of forces and feed them into the actual forces) and is done with a custom driver rather than the ASE one, so you're not restricted to the standard ASE interface nor you risk breaking other ASE-based workflows. In your case, the extra properties should be returned as an extra string, I'd say

ceriottm:
Do you want to merge this in the near future? I'd like to make a release and if you plan to work on making a MACE-based version this coming week, I'll wait.

EliaStocco:
Sorry the late reply. I am addressing this changes soon

EliaStocco:
This is the error message that I got with the `main` branch (by using a foundation model in this case):
```bash
(mace) user@hostname:~/projects/test-run$ i-pi-py_driver -m mace -u -o 
!W! THIS PES HAS NOT BEEN TESTED FOLLOWING CONVERSION TO THE NEW PES API.
/home/stoccoel/programs/miniforge/envs/mace/lib/python3.9/site-packages/e3nn/o3/_wigner.py:10: UserWarning: Environment variable TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD detected, since the`weights_only` argument was not explicitly passed to `torch.load`, forcing weights_only=False.
  _Jd, _W3j_flat, _W3j_indices = torch.load(os.path.join(os.path.dirname(__file__), 'constants.pt'))
cuequivariance or cuequivariance_torch is not available. Cuequivariance acceleration will be disabled.
/home/stoccoel/codes/mace-main/mace/calculators/mace.py:139: UserWarning: Environment variable TORCH_FORCE_NO_WEIGHTS_ONLY_LOAD detected, since the`weights_only` argument was not explicitly passed to `torch.load`, forcing weights_only=False.
  torch.load(f=model_path, map_location=device)
No dtype selected, switching to float64 to match model dtype.
Traceback (most recent call last):
  File "/home/stoccoel/codes/i-pi/bin/i-pi-py_driver", line 246, in <module>
    run_driver(
  File "/home/stoccoel/codes/i-pi/bin/i-pi-py_driver", line 126, in run_driver
    pot, force, vir, extras = driver(cell, pos)
  File "/home/stoccoel/codes/i-pi/ipi/pes/ase.py", line 84, in __call__
    properties = structure.get_properties(self.capabilities)
  File "/home/stoccoel/programs/miniforge/envs/mace/lib/python3.9/site-packages/ase/atoms.py", line 782, in get_properties
    return self._calc.calculate_properties(self, properties)
  File "/home/stoccoel/programs/miniforge/envs/mace/lib/python3.9/site-packages/ase/calculators/calculator.py", line 497, in calculate_properties
    props = self.export_properties()
  File "/home/stoccoel/programs/miniforge/envs/mace/lib/python3.9/site-packages/ase/calculators/calculator.py", line 563, in export_properties
    return Properties(self.results)
  File "/home/stoccoel/programs/miniforge/envs/mace/lib/python3.9/site-packages/ase/outputs.py", line 12, in __init__
    self._setvalue(name, value)
  File "/home/stoccoel/programs/miniforge/envs/mace/lib/python3.9/site-packages/ase/outputs.py", line 28, in _setvalue
    prop = all_outputs[name]
KeyError: 'node_energy'
```

EliaStocco:
@mahrossi @ceriottm what about adding `node_energy` to the supported properties by `ase`?
This seems to do the job.
Let me know.

ipi/pes/mace.py
```diff
@@ -38,6 +38,13 @@ def __init__(self, template, model, device="cpu", *args, **kwargs):
         except:
             raise ImportError("Couldn't load mace bindings")
 
+        try:
+            from ase.outputs import _defineprop
+
+            _defineprop("node_energy", dtype=float, shape=("natoms",))
+        except ImportError:
+            raise ValueError("Could not find or import the ASE module")
+
```
mahrossi, line 47: this looks excellent to me.

PR #435, opened 2025-04-10
Allow non-conservative forces and stresses in `metatensor` driver

frostedoyster:


ceriottm:
@venkatkapil24  I also noticed and fixed an issue with the temperature printed in NVE simulations when you don't specify an ensemble temperature. Should be uncontroversial, so if I don't hear from you I'll just merge as we need the updated driver to make PET-MAD 2x faster ^_^

ipi/pes/metatensor.py
```diff
@@ -245,7 +294,9 @@ def _compute_ensemble(positions, strain):
                     )
                 )
                 force_ensemble_tensor = -minus_force_ensemble_tensor
-                virial_ensemble_tensor = -minus_virial_ensemble_tensor
+                virial_ensemble_tensor = -minus_virial_ensemble_tensor / torch.abs(
+                    torch.det(cell)
+                )
```
frostedoyster, line 299:
```suggestion
                virial_ensemble_tensor = minus_virial_ensemble_tensor / torch.abs(
                    torch.det(cell)
                )
```

ceriottm (approved):
LGTM

PR #438, opened 2025-04-14
Remove the dependency on ase calculator for the metatensor driver

Luthaf:
This way we can evolve both calculators/drivers independently (the code in i-pi was previously relying on private details from the ASE calculator).

### DIFF
--- a/ipi/pes/metatensor.py
+++ b/ipi/pes/metatensor.py
@@ -15,8 +15,9 @@ from ipi.utils.messages import warning, info
 import ase.io
 
 torch = None
-mtt = None
+mts = None
 mta = None
+vesin_torch_metatensor = None
 
 __DRIVER_NAME__ = "metatensor"
 __DRIVER_CLASS__ = "MetatensorDriver"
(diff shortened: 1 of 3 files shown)


ipi/pes/metatensor.py
```diff
@@ -271,13 +277,11 @@ def __call__(self, cell, pos):
...
-                        types, positions @ strain, cell @ strain, pbc
+                        self._types, positions @ strain, cell @ strain, pbc
                     )
                     for options in self.model.requested_neighbor_lists():
-                        # we meed to recompute the neighbor list to be able to register gradients
-                        neighbors = mta.ase_calculator._compute_ase_neighbors(
-                            ase_atoms, options, dtype=self._dtype, device=self.device
-                        )
+                        # get the pre-computed NL, and re-attach it in the compute graph
+                        neighbors = system.get_neighbor_list(options).detach()
```
Luthaf, line 284: @frostedoyster do you have an example where `force_virial_ensemble` is used? I did not check that this code still works

Luthaf: My bad, this would be `input-ensemble.xml`

ceriottm (approved):
LGTM, I tested locally and all seems to work.

PR #444, opened 2025-05-02
Fix vesin device

frostedoyster:
For some reason, vesin can't handle non-CPU devices anymore (@Luthaf?). This new design sends the system to the GPU after the NL calculation, avoiding the bug (and a CPU-GPU-CPU trip if the bug wasn't there)


Luthaf (commented):
Changes look good, but there are conflicts wit `main`

Luthaf:
> For some reason, vesin can't handle non-CPU devices anymore (@Luthaf?)

It never handled non-CPU devices, but the previous code was based on the ASE calculator which did the same thing that's happening now.

Luthaf (approved):

PR #445, opened 2025-05-04
Fix virial in metatensor driver

frostedoyster:
The problems were due to a convention issue

### DIFF
--- a/ipi/pes/metatensor.py
+++ b/ipi/pes/metatensor.py
@@ -308,7 +308,7 @@ class MetatensorDriver(Dummy_driver):
                     .numpy(),
                 )
                 virial_ensemble = unit_to_internal(
-                    "pressure",
+                    "energy",
                     "ev/ang3",
                     virial_ensemble_tensor.detach()
                     .to(device="cpu", dtype=torch.float64)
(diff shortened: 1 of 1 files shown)


ipi/pes/metatensor.py
```diff
@@ -305,25 +299,23 @@ def _compute_ensemble(positions, strain):
...
                 force_ensemble = unit_to_internal(
                     "force",
                     "ev/ang",
                     force_ensemble_tensor.detach()
                     .to(device="cpu", dtype=torch.float64)
                     .numpy(),
                 )
-                stress_ensemble = unit_to_internal(
+                virial_ensemble = unit_to_internal(
                     "pressure",
```
ceriottm, line 311:
```suggestion
                    "energy",
```

ipi/pes/metatensor.py
```diff
@@ -305,25 +299,23 @@ def _compute_ensemble(positions, strain):
...
                     "ev/ang3",
```
ceriottm, line 312:
```suggestion
                    "electronvolt",
```

ceriottm (changes requested):
I think you forgot to change the units for the ensemble, pls confirm.

ceriottm:
I checked the "base" path and it works, we'll check the rest when we actually start using the ensembles....

ceriottm (approved):

PR #448, opened 2025-05-12
Add an interface to metatomic ML models

Luthaf:
This is duplicated and adapted from the metatensor interface.


ceriottm:
I get completely different energies than in the metatensor examples (and the trajectory blows up). Also, I didn't find instructions to run the create_model with the metatomic lj test, maybe worth adding them?

Luthaf:
> I get completely different energies than in the metatensor examples (and the trajectory blows up)

Using the test LJ model?

ceriottm (approved):
I tested it ant it works fine. We can let this coexist with the metatensor one, then deprecate and then drop it when it starts breaking.

PR #449, opened 2025-05-14
Fix "specialty" thermostats

ceriottm:
I found some lesser used thermostats were left broken by the depend object revamp. This simply fixes this and simplifies some code duplication.


litman90 (dismissed):
Looks good, please add a regtest (or at least create an issue saying a regtest is needed, so it gets done in hackathon eventually)

ceriottm:
OK I added a mega-example that combines ALL thermostats. It won't find all bugs, but it would have found this, at least it traverses the code.

Issue #454, opened 2025-05-20
Drift in conserved quantity for constrained-cell barostats


mahrossi:
As @armaet reported in #441 (see discussion there too), there is a strange drift of the conserved quantity with the MTTK barostat that needs to be investigated. Attaching most meaningful plot here. Examples in the i-PI examples folder are sufficient to reproduce.

![Image](https://github.com/user-attachments/assets/35479651-b9ed-4edf-b4f6-9c215f23b2b4)

armaet:
> Did you look carefully at the impact of traceless-ing the momentum on integrator accuracy?

I have not looked at this myself, but I can definitely ask around to see if anyone has an idea. I can also look at how this was tackled in YAFF and if this can offer any help here.

PR #455, opened 2025-05-20
Fix energy conservation in constrained barostats

ceriottm:
Addressin #454


ceriottm:
So, the tests were badly misnamed. I cleaned up the examples and now things are much clearer. Both NpT and NST have conserved quantities. The problem occurs with NPzT, or in general with cell constraints. I suspect this has somehow to do with the tracking of the conserved quantity when you zero cell velocity components, but could not figure out how.

mahrossi:
This is still WIP, right?

ceriottm:
yep. I just cleaned up the examples a bit and showed very clearly the problem. no solution in sight and I think that people wanting to use any kind of constrained barostat (including the fixed cell shape) should worry a bit about this.

mahrossi:
I will merge this even though it does not fix #454  -- please be aware of these barostat constraints as they are breaking energy conservation.

mahrossi (approved):

PR #464, opened 2025-06-17
feat/morse

EliaStocco:
A python driver that implements a Morse potential along the z axis and an harmonic potential along x and y.


mahrossi (approved):
Looks good

Issue #452, opened 2025-05-19
Python driver module not included in installation via pip


armaet:
Hi,

I noticed that, in commit [0e70ec2](https://github.com/i-pi/i-pi/commit/0e70ec235c4b2af707406923fc2a8923ed98967d#diff-1e2c670c25a308978c62213f1a0587d5663232ef25ee5d49eb6c556c5b88ee64), `setup.py` was changed such that the drivers/py module was excluded from the installation via `pip`. In our code we make use of a custom driver function and therefore we need to be able to import the `run_driver`-function itself. Why was the module excluded? Or maybe I am mistaken and it is still present somewhere?

Thanks in advance.

armaet:
Basically instead of ASE Atoms objects we use a custom class for geometries and our driver function is an adaptation of the ASE PES to handle this. Therefore it is quite convenient for us to be able to customize the driver function. 
For clarification [here](https://github.com/molmod/psiflow/blob/main/psiflow/sampling/utils.py) is our driver implementation and [here](https://github.com/molmod/psiflow/blob/main/psiflow/sampling/client.py) is where the driver is ran.

ceriottm:
If the only issue is the ASE type, I don't think there's a problem. The template is passed as a string so it's not ASE-bound. You can look at e.g. metatensor.py for a PES that doesn't inherit from ASEDriver. Am I missing something?

armaet:
Ah yes, you are correct, we could adapt to the existing ASE interface without much issue.

However, we currently set up the server and clients through custom scripts, which makes a lot of things (data staging / processing / hardware specification [CPU <> GPU]) more convenient. We also launch simulations through Parsl, so we really like to control i-PI from within python for some additional flexibility and control.

In our existing workflows, we often combine multiple PESs (e.g., a MACE model with some PLUMED bias or a harmonic approximation). We have a number of implementations used in our Psiflow package, and over time this list will surely grow. Therefore, it is useful to have a single i-PI driver that can interface with each Psiflow PES. At the moment, everything works great - provided we can access the run_driver function.

Of course we would like to operate in line with the current i-PI philosophy as much as possible. Could we achieve something similar (i.e., arbitrary combinations of PESs running on inhomogeneous hardware resources) using FFDirect?

ceriottm:
You can definitely run multiple PES on the same node and heterogeneous (CPU/GPU) using FFDirect, but you need sockets for multi-node ATM (MPI interface has been long in the making but not there in the near future). 
[...]

PR #466, opened 2025-06-25
Psiflow PES

armaet:
New driver to use FFDirect through [Psiflow](https://github.com/molmod/psiflow), as mentioned in #452. We introduced two control functions, `initialise_driver `and `check_outputs`, imported from Psiflow to improve flexibility. This avoids the need to update the driver within i-PI when we want to tweak anything.


armaet:
Indeed, FFDirect does simplify things quite a bit compared to the sockets, so it was definitely a worthwhile change for us.

>Just to understand, these two functions you import are just doing the set up and you prefer to maintain them on the Psiflow side?

Yes, it seems more convenient to keep them on our end so that we don't need to trouble you with any potential changes we might make in the future.

armaet:
I don't really know if making a minimal example through Psiflow is useful, as this would be very similar to the `examples/clients/mace` example, i.e. just a MACE model driving some system. For a standard MD simulation, Psiflow basically initializes a structure and PES (MLP, PLUMED bias, ...) and then just generates an `xml`-file which gets fed into i-PI. If it is mainly for testing purposes however, I can definitely supply a Psiflow-generated `xml`-file along with a working environment to test the Psiflow driver.

ceriottm:
OK this was mainly for you to make this more visible. If it's not important, I'll go ahead review this and merge. An interesting example of how everything can be integrated is still very much welcome in the cookbook.

armaet:
I agree, this is a good suggestion and I will look into making such an example.

PR #471, opened 2025-08-19
Set up batch evaluation on the FFDirect side

ceriottm:
This PR extends the PES API and FFDirect to bundle multiple structure evaluations in a single batched execution.
This is useful for MLIPs that can use GPUs more efficiently this way.


ceriottm:
On our side, this is ready to be merged.

EliaStocco (dismissed):

PR #481, opened 2026-01-16
Feat/python 3.12

EliaStocco:
I think the Continuous Integration with python 3.12 is broken.


mahrossi:
This was addressed, I think in #482

mahrossi (commented):
nothing here

PR #485, opened 2026-01-21
update requirements to permit numpy-2.0

ceriottm:
We really can't be stuck at numpy-1


venkatkapil24:
Screenshot 2026-01-22 at 11 42 53
interestingly, the PR page shows a green tick lol

mahrossi:
If we need to test again we try another time...

PR #487, opened 2026-01-27
Fix #477: Warn user if 'mass' is used instead of 'masses' in extxyz

### LINKED ISSUE #477: Mass input with extxyz file does not work
When I try to use the extxyz file
```
1
Lattice="1.0 0.0 0.0 0.0 1.0 0.0 0.0 0.0 1.0" pbc="F F F" Properties=species:S:1:pos:R:3:mass:R:1
H 0 0 0 2.0141
```
as input with the line `<file mode='ase'> test.extxyz </file>`, I would assume that the hydrogen atoms would get assigned a mass of 2.0141Da. And if I read the .extxyz file with ASE, it does indeed show this value in `atoms.arrays['mass']`.

However, i-PI seems to absolutely ignore this and assigns the hydrogen its usual average atomic mass. (I've checked this by outputting the density and volume of the simulation.) Why is that? [As a sidenote, manual input of the masses does work fine.]

More generally, there seems to be basically no information in the documentation about how to input different masses to i-PI (e.g., I had to look into the code to see that `units='dalton'` is needed and `units='amu'` is not recognised for some reason).

Bohaz-111:
Fixes #477.

The issue is caused by ASE's get_masses() function, which strictly looks for an array named 'masses' and ignores 'mass'. This led to silent errors where custom masses were ignored without the user knowing.

This PR adds a check to explicitly warn the user if the input file contains 'mass' but not 'masses', helping them catch this formatting error immediately. 

Please let me know if there are any further issues or require further clarification. 

Best wishes,
Bohaz

### DIFF
--- a/ipi/utils/io/backends/io_ase.py
+++ b/ipi/utils/io/backends/io_ase.py
@@ -109,9 +109,11 @@ def read_ase(filedesc):
 
     # Issue #477 fix
     if "mass" in atoms.arrays and "masses" not in atoms.arrays:
-        raise RuntimeError("ASE reader found 'mass' array but will ignore it. "
+        raise RuntimeError(
+            "ASE reader found 'mass' array but will ignore it. "
             "Rename column to 'masses' if you want to read custom atomic masses, "
-            "or remove it to use default atomic masses.")
+            "or remove it to use default atomic masses."
+        )
 
     if all(atoms.get_pbc()):
         # We want to make the cell conform


mahrossi:
Hi Bohaz, many thanks for being pro-active on this matter - I am afraid that the users will ignore the warning, though. Can you make i-PI stop with the message? And tell the user that if they want to customize the mass with this type of input file they need to change the keyword?

Bohaz-111:
Thanks Michele, looking forward to meeting you next week.

PR #489, opened 2026-01-29
Add ffrotations documentation

ceriottm:
ffrotations was not documented. I also added a brief discussion of how to include new keywords.

### DIFF
--- a/docs/src/contributing.rst
+++ b/docs/src/contributing.rst
@@ -35,7 +35,11 @@ For a PR to be issued, we will expect that:
 
 * You have produced code that is compliant with our formatting requirements (`black` and `flake8`, as explained in the README here: https://github.com/i-pi/i-pi)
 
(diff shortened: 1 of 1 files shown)


DavideTisi (approved):
works nicely!!! Maybe add the
```
The input documentation is automatically generated based on the helpstrings in the input folder.
When you add a new tag, you should also add an entry into `scripts/help.py` to have it added 
to the documentation
```
in the Contributing.md, but it is not mandatory

ceriottm:
> works nicely!!! Maybe add the
> 
> ```
> The input documentation is automatically generated based on the helpstrings in the input folder.
> When you add a new tag, you should also add an entry into `scripts/help.py` to have it added 
> to the documentation
> ```
> 
> in the Contributing.md, but it is not mandatory

True! I added it to contributors.rst in the docs because that's where the substantial info is.

PR #490, opened 2026-01-30
fixed bug within NST ensemble 

EliaStocco:
The `NST` ensemble is only physically meaningful when used with an `anisotropic` barostat (`class BaroRGB`).

Previously, the code allowed the `NST` ensemble with a flexible thermostat even when the barostat was not `anisotropic`. In this case, the pressure was left improperly initialized (set to -12345), which led to unstable and exploding simulations.

This PR enforces the use of an `anisotropic` barostat whenever the `NST` ensemble is selected. Additionally, the initial pressure is now set to `np.nan` to ensure that any improperly initialized variables trigger a hard failure. 
If a variable is not correctly initialized due to a coding error, the simulation should fail immediately rather than continue with a sentinel value like -12345.

### DIFF
--- a/ipi/utils/io/backends/io_ase.py
+++ b/ipi/utils/io/backends/io_ase.py
@@ -107,6 +107,14 @@ def read_ase(filedesc):
         _ase_open_files.pop(filedesc)
         raise EOFError()
 
+    # Issue #477 fix
+    if "mass" in atoms.arrays and "masses" not in atoms.arrays:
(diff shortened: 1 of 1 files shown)


mahrossi (commented):
👍

ceriottm (approved):
Looks good to me. We need to make it harder to do mistakes in setting up simulations and this seems a good start ^_^

PR #493, opened 2026-02-04
Multi GPU implementation for metatomic

ceriottm:
Allow to use multiple GPUs (and to oversubscribe without using batching)

### DIFF
--- a/examples/clients/metatomic/input-batched.xml
+++ b/examples/clients/metatomic/input-batched.xml
@@ -18,7 +18,8 @@
         If you have multiple gpus you can also specify device:cuda:0;cuda:1
         In this specific example, two separate threads will be started,
         and each thread will handle a batch size of 2 (that is, batch_size
-        must be greater than or equal to the number of devices requested). 
+        must be greater than or equal to the number of devices requested;
+        work will be split as evenly as possible among devices). 
     -->
     <parameters> {template:nickel.xyz,model:nickel-lj.pt,device:cpu;cpu } </parameters>
     <batch_size> 4 </batch_size>
--- a/ipi/pes/metatomic.py
+++ b/ipi/pes/metatomic.py
@@ -401,8 +401,8 @@ class MetatomicDriver(Dummy_driver):
             extras_dict["energy_uncertainty"] = list(energy_uq)
             if np.max(energy_uq) > self.uncertainty_threshold:
                 warning(
-                    f"the estimated atomic energy uncertainty {27.211*np.max(energy_uq):.4f} eV/atom "
-                    f"exceeds the selected threshold of {27.211*self.uncertainty_threshold:.4f} eV/atom",
+                    f"the estimated atomic energy uncertainty {unit_to_user('energy','electronvolt',np.max(energy_uq)):.4f} eV/atom "
+                    f"exceeds the selected threshold of {unit_to_user('energy','electronvolt',self.uncertainty_threshold):.4f} eV/atom",
                     verbosity.medium,
                 )
 


ceriottm:
@GardevoirX wait to review - I want to also try and include one small change that will use the CUDA version of vesin, so this is held back by https://github.com/Luthaf/vesin/pull/107

examples/clients/metatomic/input-batched.xml
```diff
@@ -13,8 +13,15 @@
...
 <ffdirect name='driver'>
     <pes> metatomic </pes>
-    <parameters> {template:nickel.xyz,model:nickel-lj.pt,device:cpu } </parameters>
-    <batch_size> 2 </batch_size>
+    <!-- 
+        Uses multiprocessing to run on multiple devices independently. 
+        If you have multiple gpus you can also specify device:cuda:0;cuda:1
+        In this specific example, two separate threads will be started,
+        and each thread will handle a batch size of 2 (that is, batch_size
+        must be greater than or equal to the number of devices requested).
```
GardevoirX: Or, the batches are distributed to each thread as equally as possible?

ipi/pes/metatomic.py
```diff
@@ -222,10 +269,14 @@ def check_parameters(self):
...
+        """Prepares a metatomic System object from cell and positions."""
+        if model is None:
+            model = self.model
+        if device is None:
+            device = self.device
 
-    def _prepare_system(self, cell, pos):
+        # Convert units to Angstrom
         positions = unit_to_user("length", "angstrom", pos)
         cell = unit_to_user("length", "angstrom", cell.T)
```
GardevoirX: Hmm why we take it for granted that the unit is Angstrom here, instead of taking the unit from things like `ModelCapabilities`?

ceriottm:
@GardevoirX maybe it's best to merge it now so we make a 3.1.8 release and we add later the CUDA evaluation of the NL

ipi/pes/metatomic.py
```diff
@@ -320,144 +374,178 @@ def _process_outputs(self, outputs, systems, strains):
...
                 )
-                force_ensemble_tensors, virial_ensemble_tensors = (
-                    [-j for j in jacobians[: len(systems)]],
-                    [-j for j in jacobians[len(systems) :]],
+                for force_ensemble_tensor in force_ensemble_tensors
+            ]
+            virial_ensembles = [
+                unit_to_internal(
+                    "energy",
+                    "electronvolt",
```
GardevoirX: And here

ipi/pes/metatomic.py
```diff
@@ -139,42 +158,70 @@ def check_parameters(self):
...
+            model = mta.load_atomistic_model(
+                model_path, extensions_directory=self.extensions
+            )
+            model = model.to(device)
+            self.models.append(model)
+
+        # ensure self.device and self.modelare initialized (to the first device)
+        # to ensure compatibility with single-device methods
+        self.model = self.models[0]
+        self.device = self.devices[0]
```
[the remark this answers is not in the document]
ceriottm: I don't think so. It'd be kind of crazy, but this can in principle work using mixed devices and I wouldn't limit the user from doing strange things.

ipi/pes/metatomic.py
ceriottm: we set model outputs to use ASE units, so here we are just being consistent and converting into i-PI units. you are right we could set model inputs and outputs to i-PI units so the conversion is done by the metatomiccalculator but I fear the risk of introducing bugs is high enough that we should do the change (if ever) as a separate PR

ipi/pes/metatomic.py
ceriottm: as above

ipi/pes/metatomic.py
GardevoirX: Sure we do a separate PR

GardevoirX: Sure we do a separate PR

ceriottm:
> LGTM at the code level, if urgent we can skip performance for now and merge

Not THAT urgent, but I have checked performance and it's pretty decent, for some of the systems we're working it goes almost perfectly 2x faster if you use 2 devices.

GardevoirX (approved):

PR #494, opened 2026-02-05
Custom definition of properties in the scripting module

ceriottm:
This allows defining new properties, that can then be computed and output with the usual i-PI output mechanism.

### DIFF
--- a/ipi_tests/regression_tests/tests/TEMPERATURE/nvt_langevin_md/input.xml
+++ b/ipi_tests/regression_tests/tests/TEMPERATURE/nvt_langevin_md/input.xml
@@ -17,7 +17,7 @@
       <velocities mode='thermal' units='kelvin'> 500 </velocities>
     </initialize>
     <forces>
-	<force forcefield='driver' weight='0'> </force>
+	<force forcefield='driver' weight='1'> </force>
     </forces>
     <motion mode='dynamics'>
       <dynamics mode='nvt'>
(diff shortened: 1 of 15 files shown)


ceriottm:
> Could this be made to work with compute functions provided by someone else than the user running the simulation? I.e. can we write `ipi.metatomic.compute_properties` once, and have it usable by multiple people?
> 
> Other wishes I have would be to allow using this from the `xml` input directly, and allow callable classes that take arguments for initialization. (Unless this is intended to only be accessible from the scripting API!)
> 
> ```python
> # in ipi/metatomic.py
> class compute_properties:
>     def __init__(self, template, outputs, …):
>         ...
>     
>     def __call__(self, simulation):
>         ...
> ```
> 
> and then in something like
> 
> ```
> <output prefix='script-c-p'>
>     <custom_property name="custom" class="ipi.metatomic.compute_properties"> 
>         template.xyz, outputs_name, … </def>
>     </custom_property>
>     <properties stride='2' filename='out'>[ step, custom ]</properties>
> </output>
> ```

It already works roughly like that, look at the example in `examples/features/scripting-api/run-custom-properties.py`. we could have a custom class and then one could use it in their input - only they need to choose how to call it in the constructor call. But yes, this is meant to be something to be used from the scripting API, otherwise we'd have to change a lot the scope of what a PES is.

GardevoirX (approved):
LGTM, this works well with the additional inputs recently added in `metatomic`. Maybe worth adding a more detailed example in atomistic cookbook after this [PR](https://github.com/metatensor/metatomic/pull/144) in `metatomic` repo

PR #518, opened 2026-03-26
Enable CavMD simulations of linear and multidimensional spectroscopy with accurate dipole surfaces

Astolfo-Official:
Dear Yair,

In a recent collaboration work (https://arxiv.org/abs/2603.24521), we have provided a minor code modification of the CavMD feature in i-PI to enable the concurrent connection of a molecular driver (to provide nuclear forces) and a dipole surface driver (to provide dipole moments & dipole derivatives). 

This modification not only allows a more accurate calculation of linear IR spectrum under vibrational strong coupling, but also permits the direct CavMD simulations of 2D IR-Raman spectroscopy under vibrational strong coupling, as discussed in the mentioned arXiv paper. 

We also provide an additional example of CavMD regarding how to use this new feature to compute linear IR spectroscopy under vibrational strong coupling. 

The modifications in the dipole driver and demos/2D-IR-Raman are minor. These modifications enable the direct use of the  current code for simulating vibrational strong coupling, which do not impact the original 2D-IR-Raman demonstrative calculation.

Best,

Xinwei Ji, Tomislav Begušić, Tao E. Li


mahrossi (commented):
Hey, some smaller things, I think I need a bit more time to review all steps or @litman90 could chime in too. I am thinking that a lot of infrastructure is also done or very similar to engine/motion/driven_dynamics.py so @EliaStocco could also have a look here. We could/should think about combining some of this.

drivers/f90/driver.f90
```diff
@@ -70,7 +70,7 @@ PROGRAM DRIVER
       DOUBLE PRECISION volume
       DOUBLE PRECISION, PARAMETER :: fddx = 1.0d-5
 
-      DOUBLE PRECISION, ALLOCATABLE :: dipz_der(:, :) ! Dipole (z-component) derivative (water_dip_pol model)
+      DOUBLE PRECISION, ALLOCATABLE :: dip_der(:, :) ! Dipole (z-component) derivative (water_dip_pol model)
```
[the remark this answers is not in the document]
Astolfo-Official: Thank you for pointing this out; the variable name has been corrected.

examples/features/advanced_combinations/cav_md_liquid_water/README.md
```diff
@@ -39,3 +43,44 @@ In short, to perform CavMD simulations, we need to make **two modifications** of
...
 ```
 Here, we have 216 water molecules (648 nuclei) and two photons (labeled as **L**). Please always put the photons at the very end of the xyz file. The mass of each photon is 1.0 atomic units by default. We assume a z-direction oriented optical cavity coupled to the molecules, so the k-vector of the cavity photons is along the z-direction, and the cavity photons can polarize along either the x- or y-direction. The first photon is coupled to the x direction of the total dipole moment of water, and the second photon is coupled to the y direction. Hence, for the first photon "atom", only the x coordinate is influenced by the molecules; for the second photon "atom", only the y coordinate is influenced by the molecules. In the current implementation, only two cavity photons are supported (more advanced cavity setups will be included in an updated version of the code).
+
+## More advanced dipole surfaces
+
+A more advanced way to propagate CavMD is to use more accurate dipole surfaces. The following input file and bash script use a dipole-induced-dipole water dipole surface (`water_dip_pol`, which is supported in i-PI):
+- **input_multiple_dirvers.xml**
```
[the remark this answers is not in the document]
Astolfo-Official: Thank you for pointing this out; the filename has been changed.

ipi/engine/forcefields.py
```diff
@@ -1911,18 +1923,38 @@ def get_ph_forces(self, dx_array, dy_array):
...
+        self, dx_array, dy_array, charge_array_bath=None, dipole_der=None
+    ):
         """
         Calculate the photonic forces on nuclei from MM partial charges
 
         Args:
             dx_array: x-direction dipole array of molecular subsystems
             dy_array: y-direction dipole array of molecular subsystems
             charge_array_bath: partial charges of all atoms in a single bath
+            dipole_der: the (9*natoms) derivative of the dipole moment with respect to nuclear coordinates
```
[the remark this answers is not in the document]
Astolfo-Official: Thank you for pointing this out. "(or Born effective charges)“ has been added to the corresponding comments.

ipi/engine/forcefields.py
```diff
@@ -1855,6 +1855,12 @@ def get_ph_energy(self, dx_array, dy_array):
         Returns:
             total energy of photonic system
         """
+        # convert the input dipole array to numpy array if it is a scalar
+        if np.isscalar(dx_array):
```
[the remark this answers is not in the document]
Astolfo-Official: Thank you for this comment. This was because in the previous version of the code, the returned dipole moment from the dipole driver in "extra" has a floating number data type instead of the array (so type conversion is needed). 

To avoid confusion, I have now modified it. Now when using this function, the input argument is converted to a numpy array (instead of being converted within this function).

ipi/engine/forcefields.py
```diff
@@ -1911,18 +1923,38 @@ def get_ph_forces(self, dx_array, dy_array):
...
+
+        if charge_array_bath is not None and dipole_der is not None:
+            softexit.trigger(
+                "Both charge array and dipole derivative are provided. \
+                 The forces on nuclei from cavity photons will be calculated by considering \
+                 both contributions from partial charges and dipole derivatives."
+            )
+
+        # convert the input dipole array to numpy array if it is a scalar
+        if np.isscalar(dx_array):
```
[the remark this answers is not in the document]
Astolfo-Official: This issue has been resolved in the same manner as the comment above.

EliaStocco:
As @mahrossi mentioned, I have implemented something very similar, but much simpler. 
Therefore we could think to create a more general and flexible way in i-PI to deal with 'external field interactions'.
I need to read carefully your papers to understand how to set up a common framework.

litman90:
Hi,
Thanks for your patience.

I am looking at this PR now
First general things:

- [ ] Please split the examples with and without polarizable FF into two subfolders for clarity
- [ ] Please add a regression test (as it is done for 'standard' CAVMD ipi_tests/regression_tests/tests/CAVMD, this will ensure the long-term stability of this functionality

examples/features/advanced_combinations/cav_md_liquid_water/input_multiple_drivers.xml
```diff
@@ -0,0 +1,49 @@
...
+  </ffcavphsocket>
+  <ffcavphsocket name='dipole' mode='unix' pbc="false">
+    <address> h2o-dipole-cl-cavmd </address>
+    <latency> 1e-3 </latency>
+    <apply_photon> True </apply_photon>
+    <E0> 4e-4 </E0>
+    <omega_c units='inversecm'> 3550.0 </omega_c>
+    <charge_array> [] </charge_array>
```
litman90, line 26: Defining E0, omega_c, and charge_array separately in ffcavphsocket for the PES and dipole surface seems error-prone. If the two cavphsocket instances are given inconsistent values, the behavior could become ambiguous.

A simpler and safer design might be to add a flag to ffcavphsocket, such as dipole_surface=True or False, and use it to determine which parameters are required and which should take default values.

In that case, the code should also raise a ValueError if a dipole_surface ffcavphsocket is given additional parameters that are not supposed to be set.

ipi/engine/forcefields.py
```diff
@@ -1974,18 +1976,32 @@ def get_ph_forces(self, dx_array, dy_array):
...
         """
+        if charge_array_bath is None and dipole_der is None:
+            softexit.trigger(
+                "Both charge array and dipole derivative are None. No force will be applied on nuclei from cavity photons."
+            )
+
+        if charge_array_bath is not None and dipole_der is not None:
+            softexit.trigger(
+                "Both charge array and dipole derivative are provided. \
+                 The forces on nuclei from cavity photons will be calculated by considering \
```
litman90, line 2002: This error message should contain an explanation on how to fix the input. For example, 'please provide either dipole derivatives or charge_array.'

ipi/engine/forcefields.py
```diff
@@ -2244,43 +2325,92 @@ def queue(self, atoms, cell, reqid=-1):
...
-                != int(len(pbcpos) // 3)
-            ):
-                softexit.trigger(
-                    "Total number of photons + molecules does not match total number of particles"
+
+            # this path is for the driver that is coupled to the photons, and we need to calculate photonic contributions to energy and forces
+            if evaluate_driver:
+
+                # first determine whether the dipole info is calculated using charge_array or dipole surfaces (using dipole driver)
+                has_dipole_der = ("dipole" in extra) and ("dipole_derivative" in extra)
```
litman90, line 2355: If the driver contains only dipole or only dipole_derivative, that should probably be treated as an error. In such cases, the code should fail explicitly with a clear error message.

ipi/engine/forcefields.py
```diff
@@ -2244,43 +2325,92 @@ def queue(self, atoms, cell, reqid=-1):
...
-            )
-            # check the size of photon modes + molecules to match the total number of particles
-            if (
-                self.ph.n_photon + self.n_independent_bath * self.charge_array.size
-                != int(len(pbcpos) // 3)
-            ):
-                softexit.trigger(
-                    "Total number of photons + molecules does not match total number of particles"
+
+            # this path is for the driver that is coupled to the photons, and we need to calculate photonic contributions to energy and forces
```
litman90, line 2351: As mentioned above this could be determined by a flag in the ffcavsocket, so this would become redundant

litman90 (commented):
I left a few suggestions to simplify the code and to distinguish the use of ffcavsocket for the PES versus for dipole/dipole_der with a single flag.

@mahrossi @EliaStocco I can see the overlap with your recent implementations, but because of the presence of photon “particles,” I think it is simpler to keep this handling of dipole/dipole_der separate and contained within the ffcavsocket infraestructure, at least for the foreseeable future. Would you be ok with that?

EliaStocco:
I think a common framework will be great, but the more I think about it the more I realize that it will be pretty complicated. Maybe we keep these features separated for the moment, and after I will have finished a PR on this topic (that might be helpful for cavities as well) we can think to merge them together. It will not be ready really soon, tough.

mahrossi:
Hey @Astolfo-Official - whenever you can address @litman90 comments, we would be ready to merge

examples/features/advanced_combinations/cav_md_liquid_water/input_multiple_drivers.xml
Astolfo-Official: Dear Litman,

Thank you for your valuable suggestions. I sincerely apologize for my delayed response to your comments.

I have revised the `input` and `engine` sections of the `forcefield` by adding two parameters:

1. `evaluate_photon`, which determines which driver is used to calculate the additional cavity force.
2. `dipole_surface`, which determines whether to read the dipoles and dipole derivatives from `extra` for the cavity force calculation.

By default, if these two parameters are not set, the CavMD will work as previous version without my modification.

Besides the above, I split the example and added the regression test of the new feature.

Best regards,

Xinwei Ji.

ipi/engine/forcefields.py
Astolfo-Official: Thank you for your valuable suggestions. Considering that in the main evaluation part of the program (after `if self.apply_photon`:), we already distinguish and check whether information from the dipole surface is used, only one of `dipole_derivate` and `charge_array_bath` will be passed to `get_nuc_cav_forces()`. Therefore, I removed this check and added the relevant comments.

ipi/engine/forcefields.py
Astolfo-Official: Thank you for your valuable suggestions. I have added the relevant error checks.

ipi/engine/forcefields.py
Astolfo-Official: Thank you for raising this question. When the program reaches this point, it means that we are preparing to use `charge_array_bath` to calculate the cavity force. Under the new code framework, there is no other additional check to ensure that the size of `charge_array_bath` is appropriate, so I have kept this check.

Astolfo-Official:
> Hey @Astolfo-Official - whenever you can address @litman90 comments, we would be ready to merge

Hi Mariana,

I'm very sorry for my late reply. I have already modified the code as requested by @litman90. Thank you for your efforts!

Best wishes,

Xinwei Ji.

mahrossi:
Hey Xinwei, the regtest is failing. Maybe run them locally before pushing to see if they work as intended? have a look at the README in `i-pi/ipi_tests` and you should be able to do it locally

Astolfo-Official:
> Hey Xinwei, the regtest is failing. Maybe run them locally before pushing to see if they work as intended? have a look at the README in `i-pi/ipi_tests` and you should be able to do it locally

Hi Mariana,

I ran the test locally yesterday, and it worked. I think the test file I used may have been too large, so I’ve uploaded a smaller one.

Best wishes,

Xinwei Ji.

mahrossi:
alright, I am updating the branch and if all tests run we will merge.

mahrossi:
Hi @Astolfo-Official , you have more problems in the regtests. Your file for the settings is not consistent with the input. This is what I can see, which hsould not work because you have the same address for two unix sockets and they also don't match the input you ahve in the regtest folder. I am not sure how this could run locally? that is in your "2" regtest folder. 

driver_model qtip4pf
address localhost
port 32342
socket_mode unix

driver_model water_dip_pol
address localhost
port 32343
socket_mode unix
flags -o 0

ceriottm:
Just commenting that we're preparing to release 3.2. If you want this to make it into that release, pls let us know, and iterate rapidly to solve the outstanding issues so this can be merged. Thanks!

Astolfo-Official:
> Hi @Astolfo-Official , you have more problems in the regtests. Your file for the settings is not consistent with the input. This is what I can see, which hsould not work because you have the same address for two unix sockets and they also don't match the input you ahve in the regtest folder. I am not sure how this could run locally? that is in your "2" regtest folder.
> 
> driver_model qtip4pf address localhost port 32342 socket_mode unix
> 
> driver_model water_dip_pol address localhost port 32343 socket_mode unix flags -o 0

Hi @mahrossi ,I am very sorry for the inconvenience. I have fixed the address issue and successfully run the tests on different machines. As for why the tests passed on my local machine, it might be because line 219 in ``test_tools.py`` automatically modifies the MXL address when running the tests.

litman90:
Thanks, Xinwei, for implementing these changes. @mahrossi, this is good to go for me.

Astolfo-Official:
> Just commenting that we're preparing to release 3.2. If you want this to make it into that release, pls let us know, and iterate rapidly to solve the outstanding issues so this can be merged. Thanks!

Hi @ceriottm , thanks for the heads-up. We’ve already addressed the issues, and this should now be ready for final review/merge. We’d appreciate it being considered for the 3.2 release.

Best wishes,

Xinwei Ji.

mahrossi (approved):

PR #526, opened 2026-04-02
Improve batched and multi-threaded execution logic

ceriottm:
Allow batched evaluation to work in workflows with variable numbers of evaluations (e.g. with atom swaps)

### DIFF
--- a/ipi/engine/forcefields.py
+++ b/ipi/engine/forcefields.py
@@ -544,7 +544,13 @@ class FFDirect(ForceField):
         request["t_finished"] = time.time()
 
     def launch_batch(self):
-        """Dispatches the current batch for evaluation, even if not full."""
+        """Dispatches the current batch for evaluation."""
+
+        info(
+            f"Launching batch evaluation, "
+            f"{len(self.request_batch)} / {self.batch_size}",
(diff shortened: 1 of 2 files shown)


ipi/engine/forcefields.py
```diff
@@ -1971,7 +2020,7 @@ class FFCavPhSocket(FFSocket):
 
     def __init__(
         self,
-        latency=1.0,
+        latency=1e-4,
```
mahrossi, line 2023: I am surprised this was latency was initializing as 1.0 for all ffs! This is certainly a needed improvement. I will trust that the rest of the implementation works as intended, but I can't see how it can harm other parts of the code, so I will approve this.

mahrossi (approved):

ipi/engine/forcefields.py
ceriottm: actually this was initialized to 1e-4 through the defaults in the input, but I thought it was confusing to have the function default set to 1.0 so I changed it as well.  the change COULD have non-negligible implications on other simulations, but I think only in the same corner cases where there was a problem to be fixed.
I'll merge and keep fingers crossed

Issue #509, opened 2026-03-03
Do not hardcode absolute tolerance for Born effective charge acoustic sum rule


PicoCentauri:
There is a very useful check if the Born effective charges (BEC) returned by a driver fulfill the acoustic sum rule:

 https://github.com/i-pi/i-pi/blob/a7dac15e2daa540c5533f610381cb8843a0d1d32/ipi/engine/motion/driven_dynamics.py#L274-L283

However, the absolute tolereance for the check is hardcoded as a class property:

https://github.com/i-pi/i-pi/blob/a7dac15e2daa540c5533f610381cb8843a0d1d32/ipi/engine/motion/driven_dynamics.py#L198

This is not ideal as ML models might not reach this requirement. Therefore, I suggest making this a parameter the user can vary in the input file or make it a warning.

EliaStocco:
> @EliaStocco can you address?

I have already prepared a PR to addess it.
Almost done ...

mahrossi:
Thanks! closed by #528

PR #528, opened 2026-04-13
Acoustic Sum Rule for Born Effective Charges

EliaStocco:
Letting the user to choose the tolerance.

### DIFF
--- a/ipi/engine/motion/driven_dynamics.py
+++ b/ipi/engine/motion/driven_dynamics.py
@@ -49,6 +49,7 @@ class DrivenDynamics(Dynamics):
         self,
         efield=None,
         bec=None,
+        asr_threshold=1e-8,
         *argc,
         **argv,
     ):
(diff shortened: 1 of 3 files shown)


ipi/engine/motion/driven_dynamics.py
```diff
@@ -195,21 +200,23 @@ class BEC:
     # The BEC tensors Z^* can be given to i-PI by an external driver through the extras strings in the forces
     # or they can be kept fixed during the dynamics: in this case you can provide them through a txt file.
 
-    ASR_THRESHOLD = 1e-8
+    asr_threshold = 1e-8
```
mahrossi: is this needed? I think it should be deleted

mahrossi (changes requested):
great, I think only deletion of one line as above and everything should be merged!!!

mahrossi:
addressing #509

ipi/engine/motion/driven_dynamics.py
mahrossi: I will delete and if it passes regtests, merge

mahrossi (approved):

PR #529, opened 2026-04-15
Client for PIP potentials

szquchen:
Added a client for permutationally invariant polynomial potentials under examples/clients. An example input file, the initial geometry, and a README file are provided.


ceriottm:
Thanks, @szquchen ! I had a quick look and have a suggestion that might also require small changes in `pippes`. Usually we require the user to copy the coefficient files in the folder where i-PI is run. This avoids having to modify the driver code and recompile, that is cumbersome and will put off many users. Makes sense?

szquchen:
Thank you @ceriottm for the suggestion. It makes sense. So I'll make this small change in `pippes` and the README in `examples/clients/pippes`.

ceriottm (approved):
Looks good to me. Didn't have the time to test that it actually runs, but that's more a test of PIPPES than a test of i-PI. As an example for i-PI, this is fine to merge!

mahrossi:
Thanks everyone! got here too late but fully agree! glad to see this here :-)

PR #530, opened 2026-04-15
Reorganize the scripting API as a separate module

ceriottm:
This PR creates a separate module for scripting related stuff. This is IMO useful on many levels: more visibility, more space to organize what will probably be a growing part of the codebase. Keep backward compatibility for now, and also add a couple of tests


Luthaf:
This is mainly moving code from `ipi.utils.scripting` to `ipi.scripting`, right? Then it looks fine to me, but I'm not sure I'm the best reviewer for this code =)

ceriottm:
> This is mainly moving code from `ipi.utils.scripting` to `ipi.scripting`, right? Then it looks fine to me, but I'm not sure I'm the best reviewer for this code =)

Yeah was just looking for a "OMG don't do this!" reaction if it was obviously stupid. If it's not we can look more in depth from the PoV of an i-pi user

Luthaf:
I mean, there is a re-export with warning from `ipi.utils`, so this is exactly what I would have done to move the code!

mahrossi:
This looks good to me as it is just moving. Calling attention of @hbertschi here so she knows of the move.

mahrossi (approved):

PR #531, opened 2026-04-17
Speed up the depend object and make depend_array composition based

ceriottm:
This  makes the depend machinery a bit faster, but more importantly changes it from inheriting from ndarray to using an array as member - which should simplify later on to use arbitrary classes for storage.


ipi/utils/depend.py
```diff
@@ -154,178 +103,140 @@ def __init__(
...
+        For synchronized peers, `_func` is a dict keyed by peer name; the
+        value for the manually-set peer is invoked.
         """
-
         if self._synchro is not None:
-            if not self._name == self._synchro.manual:
+            if self._name != self._synchro.manual:
                 self.set(self._func[self._synchro.manual](), manual=False)
-                info(f" @depend: Set value for {self._name} (syncro)", verbosity.debug)
+                # info(f" @depend: Set value for {self._name} (syncro)", verbosity.debug)
```
mahrossi, line 198: why are these debug outputs now all commented out? aren't they useful?

mahrossi (commented):
So this is a mega rehash of depend.py. I can see how it makes a few parts much simpler and many descriptions a lot clearer. However, I am a bit nervous to merge this without knowing which tests have been made and how to test them right. I think counting only on our regtests for such deep changes is a bit dangerous. I suppose that timings have also been checked? Which other checks have been made?

ipi/utils/depend.py
ceriottm: they are useful but quite expensive, because even when debug is disabled the f-string has to be evaluated.
The alternative would be to add an if statement, but it's still one if. these in practice are only useful to debug the depend mechanism so I think it's fine to leave them commented, rather than pay a small but noticeable cost even when running without debug outputs

ceriottm:
This passes all tests (as you can see) and indeed it is noticeably faster.  As I mention in the PR message, the main motivation however is that it makes it possible to port i-PI to torch and possibly jax, because it avoids inheriting from ndarray.

mahrossi:
alright, bear with me and I will run some "real" calculations too instead of only the reg tests for peace of mind and then I merge

mahrossi (dismissed):
after some refactoring and catching a possible bug with the help of copilot this can be approved.

mahrossi (dismissed):

PR #532, opened 2026-04-21
batched MACE

EliaStocco:
New MACE driver:

- implemented a `MACE` driver that can evaluate multiple structures in a single batch (if used via `FFDirect`).

- the class `BatchedMACE` does not depend any more on `mace.calculators.MACECalculator` apart from the initialization.

- the file` ipi/pes/_mace.py` works also as a standalone script.


mahrossi:
@EliaStocco could we walk through this together? :-) will speed up the review process

mahrossi:
Hi there, there is a big conflict now with the main branch - could you explain what was done (and why it was needed) in ase.py?

mahrossi:
I solved the conflict but would like confirmation that calculations run as expected in this branch before merging

mahrossi (dismissed):

drivers/py/driver.py
```diff
@@ -101,7 +101,7 @@ def run_driver(
             # initialization
             rid = recv_data(sock, np.int32())
             initlen = recv_data(sock, np.int32())
-            initstr = recv_data(sock, np.chararray(initlen))
+            initstr = recv_data(sock, np.empty(initlen, dtype="S1"))
```
mahrossi, line 104: Are we sure we always expect a 1 byte string?

ipi/pes/_ase.py
```diff
@@ -35,14 +35,6 @@ def __init__(
         *args,
         **kwargs,
     ):
-        global read
```
mahrossi, line 38: why was this warning removed?

ipi/pes/_ase.py
```diff
@@ -2,12 +2,12 @@
 
 import json
 import numpy as np
-from .dummy import Dummy_driver
+from ipi.pes.dummy import Dummy_driver
 
 from ipi.utils.units import unit_to_internal, unit_to_user
-from ipi.utils.messages import warning
 
-read = None
```
mahrossi, line 10: this was weird, I think we really need the new call.

ipi/pes/_ase.py
mahrossi, line 38: Why were the warnings below removed?

ipi/pes/_ase.py
```diff
@@ -35,14 +35,6 @@ def __init__(
         *args,
         **kwargs,
     ):
```
mahrossi, line 53: In the old code, there were import warnings that seem to have been removed. Isn't this change dangerous? it is good if the user knows why things crashed

ipi/pes/_ase.py
```diff
@@ -2,12 +2,12 @@
...
 import json
 import numpy as np
-from .dummy import Dummy_driver
+from ipi.pes.dummy import Dummy_driver
 
 from ipi.utils.units import unit_to_internal, unit_to_user
-from ipi.utils.messages import warning
 
-read = None
+from ase.io import read
```
mahrossi, line 25: if ASE not installed this will throw an error.

ipi/pes/_ase.py
```diff
@@ -2,12 +2,12 @@
...
+from ase import Atoms
```
mahrossi, line 26: same here. what was the problem with the old code that imports inside the if clause and informs the user if this is not installed? otherwise i-PI can crash for everyone that does not have ASE.

drivers/py/driver.py
EliaStocco: I think I have modified it because I was getting some warnings from `numpy`. 
`chararray` has indeed been [deprecated](https://numpy.org/devdocs/reference/generated/numpy.char.chararray.html)

ipi/pes/_ase.py
EliaStocco: Right now every file in `ipi/pes` gets executed only if strictly necessary. 
`_mace.py` will be executed only if the user specifies "mace" in the `input.xml` or via `i-pi_py_driver.py`.
The same holds for all other files, like `_ase.py`
Therefore it is safe to directly import `ase` at the beginning of the file.
If the user specifies a client that requires `ase` ... then they should install it.
I do not think that `warning("Could not find or import the ASE module")` is better than the exception automatically raised by python if `ase` is not found.

ipi/pes/_ase.py
EliaStocco: It does not anymore, since this [PR](https://github.com/i-pi/i-pi/pull/472)

drivers/py/driver.py
mahrossi: yep, just asking if it is always dtype="S1" or we should leave it more flexible.

ipi/pes/_ase.py
mahrossi: I think the python exceptions sometimes do not get noticed by users. It is a better user experience if you inform them what happened. This is a simple thing to do to make users more at ease with this if ase is not in requirements.txt of i-PI

ceriottm:
FWIW, mosrt of the other PES do delayed imports, even if the files are not imported unless used.

mahrossi (approved):

PR #533, opened 2026-04-28
Allows constrained atoms with svr on the centroid and fixes conserved quantities with fixed atoms in CMD setups

mahrossi:
This PR adds the possibility of fixing atoms with a global thermostat on the centroid and fixes a bug in calculating conserved quantity with CMD and fixed atoms, both when using local and global thermostats.

I know this is not a general fix and we are not doing real constrained dynamics on the centroid - which could be the more physical way of implementing this. I am planning another PR with a proposition for that. However, this is a practical tool for many simulations and not difficult to enable.


ceriottm (dismissed):
I fixed what I think was a bug. Redy to merge if you agree.

PR #536, opened 2026-05-01
Some updates of the metatomic module

GardevoirX:
1. Applied [the neighbor list calculator caching](https://github.com/Luthaf/vesin/commit/2ce4ca111372e65610e2109f62cc4ce3d242da25#diff-1ed1a7848e2057fb156fa15a63c126283d1bdb9150240d396c7af12e72a563d5) in i-pi, as discussed before
2. No longer raise an error when the version of metatomic-torch isn't v0.1, the same behavior as what introduced in https://github.com/Luthaf/vesin/commit/351904f740a88da610b3382365db042dc394544e


ceriottm (commented):
LGTM, and I tested it works well locally (could not see speedup because CPU MLIPs are too slow to notice)

ceriottm (approved):

PR #539, opened 2026-05-08
Convert i-PI to use array_api backend

ceriottm:
This is a MASSIVE PR, and it'll probably leave a lot of stuff broken. It uses array_api and array_api_compat to make the whole of i-PI able to use numpy or torch as array backends, interchangeably.

The backend can be selected with an environment variable, or a command-line parameter. It should be relatively straightforward to include also cupy or jax, but I'd first get a fully-functional torch version first. 

Some design considerations:
* All of i-PI functions in the array_api layer, except for the sockets (that use numpy arrays) and a few math functions that are numpy-specific (that we could convert later, but are of marginal importance)
* This requires calling dstrip *in every instance*, to avoid ambiguity

TODO before merging:
- [ ] Discuss the design and possible interactions with other software
- [ ] Test as thoroughly as possible, on both CPU and GPU
- [ ] Verify the speed of some basic workflows on main, and on this branch using numpy, torch.cpu and torch.gpu


ceriottm:
Closing this and working on the i-pi-cosmo fork

PR #540, opened 2026-05-12
enh/clean exit

venkatkapil24:
This introduces as simulation.finished bool so that softexit is triggered only once. Currently, it is introduced in simulation.step() for geop, phonons, etc. but not in the same way for MD. This is a step towards easier scripting and codecoverge integration.


venkatkapil24:
one step towards #541

bin/i-pi
ceriottm, line 1: Why all these formatting changes? Are you on the same black version?

ipi/engine/motion/instanton.py
```diff
@@ -19,9 +19,8 @@
 
 from ipi.engine.beads import Beads
 from ipi.engine.normalmodes import NormalModes
-from ipi.engine.motion import Motion
+from ipi.engine.motion import Motion, MotionExit
```
ceriottm: I don't think there's a need for a custom exception. It was wrong in the first place to call softexit for a error in the options. We should call finish() when the simulation runs out, like in geop, but if there's an error we don't need a custom exception to kill the run.

ipi/engine/motion/motion.py
```diff
@@ -9,6 +9,15 @@
 from ipi.utils.depend import depend_value, dproperties
 
 
+class MotionExit(Exception):
```
ceriottm: I think this custom exception should go.

ipi/engine/simulation.py
```diff
@@ -427,28 +431,73 @@ def run(self, write_outputs=True):
 
         self.rollback = False
 
+    def _check_finished_motions(self):
```
ceriottm: This seems overly complicated - if you're finishing a motion class, then what's the point of checking if any of the motion classes is finished? of course _check_finished_motions will return true!

ipi/engine/simulation.py
```diff
@@ -427,28 +431,73 @@ def run(self, write_outputs=True):
...
-                stepthreads.append(st)
+                stepthreads.append((st, s.motion))
 
-            for st in stepthreads:
+            motion_finished = False
+            for st, motion in stepthreads:
                 if softexit.triggered:
                     return
-                st.result()
+                try:
```
ceriottm: I can't see a reason to have an exception here - again over-engineering, I'd get rid of the MotionExit machinery

ipi_tests/unit_tests/engine/test_motion_exit.py
ceriottm, line 1: looks reasonable, again modulo MotionExit

ceriottm (changes requested):
Looks good overall, but this MotionExit custom exception does not convince me.

bin/i-pi
venkatkapil24: likely

ceriottm (dismissed):
Check that restarts work well, that there are no discontinuities or the like when you touch EXIT and restart from the RESTART file. Otherwise lgtm

venkatkapil24:
Good shout. Looks like there are some issues with restarts (not for MD). WIP.

venkatkapil24:
> Good shout. Looks like there are some issues with restarts (not for MD). WIP.

FYI, the "issues" were just numerics as the RESTART files change the precision w.r.t. variables.

ceriottm (approved):

PR #543, opened 2026-05-16
Allow for more readable file type in `FFPlumed`

GardevoirX:
`read_file` is used in initializing `FFPlumed`. It supports `xyz`, `pdb`, `ase`, and `chk`, while currently we only read when the mode is `xyz`. This PR allows for reading when the mode is `pdb` or `ase`, and throwing an error when the mode isn't `xyz`, `pdb`, or `ase` 

I didn't add `chk` because I know little about it, and I don't see why ppl would pass the `chk` file as the initial structure.


ceriottm (approved):
Can't recall why this was just limited to xyz.

PR #547, opened 2026-05-26
Fix asr threshold computation

PicoCentauri:
The class property `BEC.asr_threshold` does not exist. The actual property that we want is from the namespace of the current object.


mahrossi:
this is right. @EliaStocco can you check what happened when you fixed this last? I will accept this.

mahrossi (approved):

PR #549, opened 2026-06-12
Add batching to the socket interface

ceriottm:
Add the same inline batching mechanism used for ffdirect to ffsocket. 
Batch size is communicated to the client on initialization, and structures are collected and dispatched together in a single message, then unpacked and returned as multiple force/energy/stress blocks.


mahrossi (approved):
This has been reviewed, tested and made slightly more robust. I have not tested speed specifically, but the batched socket communication works.

PR #551, opened 2026-06-14
SHM communication

ceriottm:
Enable communication using shared memory. ~15% faster than unix-domain socket

Should be merged after #549, as it is based upon it


ceriottm:
Closing in favor of #553 that also includes MPI communicators and rationalizes the whole communication machinery

PR #552, opened 2026-06-16
Enh/regtest cov integration

venkatkapil24:
1. Route test_run.py regtests through InteractiveSimulation so i-PI runs inside pytest and contributes to coverage.
2. Add a run_ipi_in_process switch to the regression runner.
3. Reset global softexit test state after in-process regtest runs.


venkatkapil24:
okay, so pytests are now in-process. I will now add changes to the CLI which may break the tests. Please note that the tests work up to this point.

venkatkapil24:
The tests are failing expectedly due to inaccess to codecov token. The next step will be someone with login credentials @ceriottm @mahrossi needs to setup codecov with i-pi following 

https://docs.codecov.com/docs/quick-start

I have made the necessary changes to the CI.

ceriottm:
Have activated codecov integration (I hope)

venkatkapil24:
I am not sure if I agree with the copilot change. That does "fix" the failing tests by disabling CI to fail if codecov doesn't work lol. I will revert back and try again.

venkatkapil24:
so the codecov token has to be added to secrets 

https://docs.codecov.com/docs/adding-the-codecov-token
Screenshot 2026-06-19 at 17 46 18

ceriottm:
it has been added already 
image

mahrossi:
I now re-ran the tests and it does seem to have worked!

ipi_tests/test_tools.py
```diff
@@ -77,6 +84,105 @@ def clean_tmp_dir():
             pass
 
 
+class InProcessIPI:
```
mahrossi, line 87: I would just like to see a more verbose documentation of this class if possible?

mahrossi (approved):
I think this is really great. Just a better documentation of the new class would be helpful for the future and probably straightforward to add!

venkatkapil24:
cool, what coverage do we get? :D

PR #553, opened 2026-06-20
SHM and MPI communication modes

ceriottm:
Greatly extend the ways i-PI can communicate with the drivers. Speedup of communicaation is hit-and-miss, but this opens up paths for optimization for specific use cases. 

The PR also attempts to improve the documentation around this, and to test better the various communication modes.

Supersedes #551


docs/src/distributed.rst
```diff
@@ -3,6 +3,64 @@
 Distributed execution
 =====================
 
+.. _transports:
+
+Choosing how to talk to the driver
```
mahrossi, line 8: nice!

docs/src/distributed.rst
```diff
@@ -195,6 +264,154 @@ One can specify a parameter “timeout”, that corresponds to the maximum
...
+is inexpensive; for expensive potentials it makes little difference. If fewer than
+``batch_size`` structures are queued, the socket transport pads the batch by
+replicating the last structure, so the run never stalls. The in-process
+``ffdirect`` backend does not pad; instead, when run with ``threaded='True'`` it
+flushes an incomplete final batch (evaluating fewer than ``batch_size`` structures)
+once its poll loop goes idle. Without threading there is no idle flush, so the
+number of structures evaluated must be a multiple of ``batch_size`` or the run
+hangs.
+
+Batching requires ``consolidate_messages`` (the default) and is incompatible with
```
mahrossi, line 319: thanks for this documentation, it is very clear!

docs/src/distributed.rst
```diff
@@ -195,6 +264,154 @@ One can specify a parameter “timeout”, that corresponds to the maximum
...
+ranks automatically at start-up, and the number of drivers is set purely by the
+launch command.
+
+Everything that works over sockets works over MPI: multiple drivers evaluate
+replicas in parallel, ``batch_size`` bundles several structures per request, and
+the ``extras`` string (e.g. a dipole) is returned in the same way. In addition,
+each driver may itself be an MPI-parallel program: passing ``--group-size G``
+(``-g``) to the driver splits the driver ranks into groups of ``G`` ranks, where
+only the root of each group communicates with i-PI and broadcasts the work to its
+group.
```
mahrossi, line 365: should add to a todo/hackathon list, a host of examples of commands one could run.

drivers/py/driver.py
```diff
@@ -118,6 +126,48 @@ def run_driver(
             if f_verbose:
                 print(rid, initstr, "batch_size:", batch_n)
             f_init = True  # we are initialized now
+        elif header == Message("POSDATA") and shm:
```
mahrossi, line 129: clean! nice!

examples/clients/communication/mpi/harmonic/Makefile
```diff
@@ -0,0 +1,15 @@
+# Makefile for the MPI harmonic oscillator example
```
mahrossi, line 1: I would vote for organizing this differently. We can just have a folder called "transport_options"  where we have one example with mpi, one with shm, one with direct, one with socket-inet and one with socket-unix -- what do you think?

mahrossi (commented):
I need to run some examples. It is all really good. Just one relevant question regarding organization of examples - after decision we change.

examples/clients/communication/mpi/harmonic/Makefile
ceriottm: yes I agree. good idea these are not really "clients" just a demo of communication modes

ceriottm: ok, done

mahrossi (dismissed):
nice, the examples are good. I think that "direct" could also be added just to make it prominent, but we can also open a separate PR to add that

venkatkapil24 (approved):

PR #554, opened 2026-06-23
Revert "Enh/regtest cov integration"

mahrossi:
Reverts i-pi/i-pi#552


ceriottm:
let's see if we can fix in #553

mahrossi:
yep I closed this

PR #557, opened 2026-06-26
Accessible version

mahrossi:
This grabs a feature of #419 that is very useful, even if we do not implement yet the more complex input versioning (which we are undecided about).

Now version is accessible from within the code and it is also available if i-PI is not installed from pip. Version bumps need to be changed in `ipi/__init__.py`

### DIFF
--- a/ipi/__init__.py
+++ b/ipi/__init__.py
@@ -2,7 +2,7 @@
 The i-PI module.
 """
 
-__version__ = "3.2.0"
+__version__ = "unknown"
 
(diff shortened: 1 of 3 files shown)


ceriottm:
I don't like that we now have two sources of truth about the version.

ceriottm:
Made another attempt making sure that we don't have to set the version in two different places. Check out #559

mahrossi:
But here it was only in one place

PR #559, opened 2026-06-27
Document and improve the release process

ceriottm:
Version control is now based on a single file, and release workflow checks it matches the tag

Alternative to #557


mahrossi:
IDK looks ok, I guess I will close #557  that was doing the same thing but differently.

PR #560, opened 2026-06-29
Fix a bug in parsing of output traj files

ceriottm:
Traj files with properties other than position would be read as zero. Also added tests to make sure this does not break again.


ceriottm:
Now with >50% project code coverage ^_^

mahrossi (approved):
Ha yes, this is necessary -- good. Codecov is somewhat painful. But great progress!

PR #563, opened 2026-07-08
An implementation of economized path integrals

ceriottm:
Heavily AI-assisted, after some careful planning. Checked it in depth, and everything seems to work well.

### DIFF
--- a/ipi/utils/nmtransform.py
+++ b/ipi/utils/nmtransform.py
@@ -5,6 +5,8 @@
 # See the "licenses" directory for full license information.
 
 
+from functools import lru_cache
+
 import numpy as np
 
 from ipi.utils.depend import dstrip
@@ -19,6 +21,7 @@ __all__ = [
     "mk_o_nm_matrix",
     "nm_eva",
     "o_nm_eva",
+    "eco_eva",
 ]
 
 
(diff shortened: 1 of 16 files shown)


ceriottm:
I've now checked this carefully, and looks like it works well. I think that a lot of the "eco guards" I put in place can probably be lifted (most estimators should still work and simply need using the eco frequencies instead than the trotter ones).

ipi/utils/nmtransform.py
```diff
@@ -55,6 +56,119 @@ def o_nm_eva(nbeads):
...
+def _eco_fit(nbeads, xmax, y0=None):
+    """Fits the dimensionless internal-mode parameters y_k = beta*hbar*omega_k
+    of the Eco path integral, minimizing the rms fractional error in the
+    radius of gyration of harmonic oscillators with 0 <= beta*hbar*omega <= xmax.
+    Follows the reference implementation in the supplementary material of
+    Zeng & Manolopoulos, "Economised path integrals": safe Newton iterations
+    with an eigenvalue-shifted Hessian and a line search that keeps the y_k
+    positive and in ascending order. Starts from the Matsubara frequencies,
+    or from the initial guess y0 (e.g. a previous solution) if given.
+
```
mahrossi, line 85: add citation to https://arxiv.org/abs/2607.06414

mahrossi (commented):
Could you ask Copilot or your other AI friend to add a piece of documentation on this? Just to say what it is - as maybe the future will see name changes.

ipi/utils/nmtransform.py
```diff
@@ -55,6 +56,119 @@ def o_nm_eva(nbeads):
...
+        g = (r[:, np.newaxis] * dg).mean(axis=0)
+        h = (dg.T @ dg + np.diag(r @ d2)) / m
+        return s, g, h
+
+    if y0 is not None:
+        y = np.array(y0, float)
+    else:
+        y = 2.0 * np.pi * np.arange(1, nfree + 1, dtype=float)  # Matsubara guess
+    s, g, h = objfun(y)
+    for _ in range(500):
```
litman90, line 119: Perhaps add an error message if we haven't converged the optimization with 500 steps

litman90 (changes requested):
A few comments:
- Have you run some simulations to test functionality? Until you or someone else does, I would refrain from merging. Next week I am traveling, but the week after I could do it; let me know
- I didn't find safeguards when doing instanton calculations; probably worth adding
-Same when using GLE-type thermostats, but that might be a bit trickier.

PR #564, opened 2026-07-09
Utilize the verlet list of `vesin`

GardevoirX:
Adopting the new feature introcuded in vesin 0.6: https://github.com/Luthaf/vesin/releases/tag/v0.6.0


ceriottm:
I would NOT include the test of the PES. however I'd include the skin parameter in the examples, and make sure that the LJ example runs with this locally.

GardevoirX:
> however I'd include the skin parameter in the examples, and make sure that the LJ example runs with this locally.

Included, but cannot pass the cov tests

> I would NOT include the test of the PES. 

That's why I pushed the tests. And I think if there's no test, in future no one can modify the code because they cannot pass the cov test

ceriottm:
> > however I'd include the skin parameter in the examples, and make sure that the LJ example runs with this locally.
> 
> Included, but cannot pass the cov tests
> 
> > I would NOT include the test of the PES.
> 
> That's why I pushed the tests. And I think if there's no test, in future no one can modify the code because they cannot pass the cov test

Yes, the codecov is set up in a silly way to push to increase the coverage, but does't make sense like this so do not worry I will override and merge.

ceriottm (approved):
Tested it, and it works. Couldn't verify how much it speeds things up, but it runs.

PR #567, opened 2026-07-23
batched MACE

EliaStocco:
# Improve MACE batching and add a batched example

## Summary

This PR improves the performance and configurability of the MACE interface and adds a batched example.

## Changes

- Batch all structures in `compute_batched()` by default while preserving an explicit batch-size override.
- Compute `node_e0` once per collated batch instead of once per structure.
- Print a one-time summary of model outputs and GPU-to-CPU transfers, including instructions for ignoring unneeded optional properties.
- Allow users to extend `ase_like_properties` programmatically or through the standalone CLI.
- Add `examples/clients/mace-batched` with a four-bead `<ffdirect>` calculation, a small water structure, MACE settings, model download script, documentation, and recorded output.
- Exclude the downloaded `mace.model` binary from version control.


mahrossi:
I believe this is, regardless, an improvement to the original state or? No harm to merge?

mahrossi (commented):
it is good that there is an addition of an example and some clean-up. the reason for the lack of performance is still mysterious though. that warrants keeping the issue open. however, this merge, after properly fixed for regression tests, would represent an improvement.

Issue #546, opened 2026-05-21
Is it possible to calculate isotope_zetasc in the NPT ensemble?


wxwth:
Dear i-pi developers,

I tried to calculate the direct scaled-coordinate estimator (isotope_zetasc) in the NPT ensemble (330 K, 1atm). However, as shown in the enclosed output files, the derived <exp(-beta * sc)> values seem unreasonable. When I switch to the NVT ensemble, the values become reasonable.

I am wondering whether it is, in principle, impossible to calculate isotope_zetasc in the NPT ensemble, or whether there may be any issues with the enclosed input file or with the i-pi code. Thank you very much!

Sincerely,
Xu

P.S. Could you please reopen my previous issue regarding the direct thermodynamic estimator (isotope_zetatd)? https://github.com/i-pi/i-pi/issues/470 It seems that some issues may still remain. Thank you!

[input.xml.txt](https://github.com/user-attachments/files/28096572/input.xml.txt)
[simulation.SC.txt](https://github.com/user-attachments/files/28096571/simulation.SC.txt)

mahrossi:
Hey Xu, I did reopen, but actually we are not completely sure who is maintaining this functionality. I will tag @Bohaz-111 here as he did act on the last issue. Do tag us directly so the issues get more attention! If you can find a fix, please contribute directly - it is normally the fastest way to solve an issue as your need is a priority for you. We really welcome it! See here for more details https://docs.ipi-code.org/contributing.html

Bohaz-111:
Dear Xu, 

I will take look at this and try to fix it. 

Best wishes,
Bohaz

Bohaz-111:
Dear Xu,

## isotope_zetasc in NPT
isotope_zetasc is computed correctly in principle for NPT, but the implementation was missing a cell update.
[...]

PR #574, opened 2026-09-08
Sync dummy cell before scaled-coordinate isotope estimators

Bohaz-111:
Reported in issue #546 .


Bohaz-111:
Non of the test excecuted the added lines, codecov target could not be reached

Bohaz-111:
`isotope_zetasc` returns meaningless values under NPT, while being correct under
NVT. Fixes #546.

## Cause

The scaled-coordinate estimators evaluate the perturbed configuration on a dummy
cell created by `system.cell.clone()` at bind time. `Cell.clone()` is a detached
copy, so it never tracks the real cell — while the barostat mutates `cell.h`
every inner timestep. The estimator therefore computes

sc = V(q_scaled ; h_0) - V(q ; h(t))

with a different box on each side. The error is the whole-system energy
difference between the stale and current boxes: extensive, and carrying no
isotope information. Under NVT `h(t) = h_0`, so it vanishes exactly and the bug
is invisible.

## Test

New regression test `ipi_tests/regression_tests/tests/NPT/isotope-sc-npt/`:
64 q-TIP4P/F waters, 4 beads, 30 steps, isotropic barostat at 20 kbar so the
cell compresses ~16%. Covers both the `Properties` and `Trajectories` paths.

under NVT, or with a barely moving cell, the
test could not detect the bug. 
With the fix, exp(-betaP*sc) stays ~1.0. Without it, sc grows to ~6.3 Ha,
giving an exponent of about -1670, so the estimator underflows to exactly 0 in
double precision.

mahrossi (commented):
Thanks, this is a super needed bug fix and especial extra thanks to have added reg test! Would it be ok to decrease the number of steps and frames for the reference to 10 though? This is to curb the size of the repo. If there is a reason to do 30 steps no problem, but 10 would be best in the long run.

Bohaz-111:
Yes, 10 steps is enough

venkatkapil24:
this ready? to me it looks good :)

mahrossi (approved):

PR #576, opened 2026-09-10
fixed bug with initialization

EliaStocco:
## Summary

Fix checkpoint-based initialization (`<file mode='chk'>`) so it reads only the state required by the initializer:

- beads (positions, momenta, masses, labels)
- cell
- motion state

Checkpoint force fields are no longer constructed in this path. This prevents initialization from failing when an old checkpoint contains a stale or incompatible force-field configuration --> **good for backcompatibility**.

## Design

A new lightweight `InputSimulation.fetch_checkpoint_state()` method fetches the first system’s beads, cell, and motion directly from the parsed input objects.

Normal full-checkpoint restart behavior is unchanged: `InputSimulation.fetch()` still reconstructs all force fields.

## Tests

Added regression coverage for a checkpoint containing an `ffdirect` configuration that fails when instantiated:

- full checkpoint loading still raises the expected error;
- initialization from that checkpoint succeeds;
- positions, momenta, cell units, masses, labels, and motion state are transferred correctly.

Tested with:

```bash
.tox/unit/bin/pytest -q ipi_tests/unit_tests/engine/test_initializer.py
.tox/unit/bin/pytest -q ipi_tests/unit_tests/engine
```

### DIFF
--- a/ipi/engine/initializer.py
+++ b/ipi/engine/initializer.py
@@ -146,30 +146,21 @@ def init_chk(filename):
        filename: A string giving the name of the checkpoint file to be read from.
 
     Returns:
-       A Beads object, Cell object and Thermostat object as read from the
+       A Beads object, Cell object and Motion object as read from the
        checkpoint file.
     """
 
-    # reads configuration from a checkpoint file
-    rfile = open(filename, "r")
-    xmlchk = xml_parse_file(rfile)  # Parses the file.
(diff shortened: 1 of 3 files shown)


venkatkapil24:
is this ready?

EliaStocco:
It is ready.

venkatkapil24 (approved):
Really good quality PR!

PR #577, opened 2026-09-15
Add stochastic cell-rescaling barostat

bkmi:
Implements the Bernetti–Bussi stochastic cell-rescaling barostat for classical NPT simulations.

  - Adds the stochastic-rescaling barostat mode
  - Supports configurable compressibility, relaxation time, and stride
  - Preserves state across checkpoints/restarts
  - Adds unit conversions, documentation, an example, and regression coverage
  - Updates the contributing guide to reference main

  Current limitations: one bead, OBABO splitting, and single-time-step integration.

  Validation:

  - Format and lint passed
  - 329 unit tests passed
  - 189 examples passed
  - 3 MPI tests passed
  - Documentation built successfully
  - 70/71 regression tests passed; the unrelated COMMITTEE/XTRA-JSON test timed out locally
  - SCR regression and example passed


ceriottm:
Thanks for the contribution. Are you sure it's necessary to do this with all these special cases? This barostat, as with everything Bussi, is based on an elegant Trotter splitting, and so should be implementable just as a different `Barostat` subclass that uses the same isotropic integrator, and simply calls a different qcstep with different parameters. 

Let me know if you have looked into this more deeply and see a reason why this shouldn't work. Otherwise I've two ways to go forward that I propose. (1) you ask your agent to reimplement it in a way that does not touch `dynamics.py`. (2) I ask my agent to implement it from scratch without touching dynamics.py and I let you check that PR to verify it addresses your needs. 

Thanks again, and I'll wait for you to tell me what you prefer (and/or if I missed something indeed).

bkmi:
You are totally right. I revisited the splitting and reimplemented SCR as a Barostat subclass using the existing NPT integrator, with no changes to `dynamics.py`. I also removed the stride-specific scheduling and updated the tests and reference trajectory.

thanks for the tip!

